PR #6, opened 2017-09-01
Adding TCP port 1234 to Etherbone dissector.

mithro:
LiteEth designs seem to commonly use TCP port 1234.


enjoy-digital:
Thanks.

PR #17, opened 2019-05-30
Remove call to identifier_frequency_read() function

jeanthom:
This function doesn't seem to exist anymore and causes build issues.


enjoy-digital:
I removed the libuip example code. Example of LiteETH integrated with libuip can be found in https://github.com/timvideos/HDMI2USB-litex-firmware/tree/master/firmware.

PR #26, opened 2020-01-17
A few minor changes that help RGMII phy related debugging. {s6, s7, u…

jersey99:
Having some signals instance variables helped with debugging. Synthesizers, should optimize them, so they should come at no cost (?)

Make dw in the phys class variables.

### DIFF
--- a/liteeth/core/ip.py
+++ b/liteeth/core/ip.py
@@ -208,7 +208,7 @@ class LiteEthIPRX(Module):
                 NextState("CHECK")
             )
         )
-        valid = Signal(reset_less=True)
+        self.valid = valid = Signal(reset_less=True)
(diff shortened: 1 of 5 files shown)


jersey99:
Also, would be a good time to look into having a single constructor for {s6, s7, us, ecp5}rgmii.py.  Also, when abstract components from nmigen like [this](https://github.com/m-labs/nmigen/blob/master/nmigen/vendor/xilinx_7series.py) would get more stable, we potentially merge them all into a single file.

enjoy-digital (approved):

enjoy-digital:
Thanks! Yes that's a good idea starting simplifying/regrouping the different rgmii phys. I'll try to work on that.

PR #35, opened 2020-03-17
Allow changing all SoC options through YAML config

Xiretza:
This can be useful for the likes of `csr_data_width`, for which special-casing would result in quite a messy script.

`mem_map` and `csr_map` are now children of the `soc` config tree, old configs fail with a warning instead of being silently ignored.


enjoy-digital:
Thanks, merged.

PR #36, opened 2020-03-19
mac: add crossbar for sharing PHY between HW ethernet cores and Wishbone

piotr-binkowski:
This PR adds `hybrid` mode to the LiteEthMAC which allows to share the MAC/PHY between IP cores and SoftCPU. Code might require a cleanup but in its current work its working fine in litex_sim and on real HW (NeTV2 with litex-buildenv SoC)

### DIFF
--- a/liteeth/mac/__init__.py
+++ b/liteeth/mac/__init__.py
@@ -10,7 +10,9 @@ class LiteEthMAC(Module, AutoCSR):
                  endianness="big",
                  with_preamble_crc=True,
                  nrxslots=2,
-                 ntxslots=2):
+                 ntxslots=2,
+                 cpu_dw=32,
+                 hw_mac=None):
         self.submodules.core = LiteEthMACCore(phy, dw, endianness, with_preamble_crc)
         self.csrs = []
         if interface == "crossbar":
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks a lot. I'm merging and will do eventual cleanup/changes very soon.

liteeth/mac/__init__.py
```diff
@@ -31,8 +33,157 @@ def __init__(self, phy, dw,
...
         return self.csrs
+
+class LiteEthMACCoreCrossbar(Module):
+    def __init__(self, core, crossbar, interface, dw, cpu_dw, endianness, hw_mac=None):
+        fifo_depth = 2048
+        wishbone_rx_fifo = stream.SyncFIFO(eth_phy_description(dw), fifo_depth, buffered=True)
+        wishbone_tx_fifo = stream.SyncFIFO(eth_phy_description(dw), fifo_depth, buffered=True)
+        crossbar_rx_fifo = stream.SyncFIFO(eth_phy_description(dw), fifo_depth, buffered=True)
+        crossbar_tx_fifo = stream.SyncFIFO(eth_phy_description(dw), fifo_depth, buffered=True)
+
```
enjoy-digital, line 62: @piotr-binkowski: i'm just having a closer look at the code, can you explain the purpose of these FIFOs? I just want to see if we can avoid them since they are relatively large and will use Block RAMs.

piotr-binkowski: I think that those FIFOs are just a leftover from my early attempts at making the crossbar and removing them shouldn't break anything. You might want to check if it still works litex_sim as I had some issues with Verilator when using hw_mac configuration that were solved by adding small FIFOs there

enjoy-digital: Ok thanks, i'll look at that.

PR #42, opened 2020-06-25
mac padding: fix counter reset value

shuffle2:
I observed following behavior:
 1. first sent packet after the module reset required padding
 1. incorrect reset value would cause 1 fewer padding bytes to be appended than were required
 1. too-short packet was dropped as invalid by remote receiver


enjoy-digital:
Thanks for catching and fixing this issue!

PR #51, opened 2020-10-11
phy/ecp5rgmii: Fix io delay blocks

### LINKED ISSUE #50: phy/ecp5rgmii: Incorrect parameters in DELAYF blocks
In some recent debugging I've discovered that nextpnr does not consider strings such as `"DELAY80"` a valid parameter value for `DEL_VALUE` 

See in this snippet, taken from the ecp5rgmii phy

```python
rx_delay_taps = 80

self.specials += [
       Instance("DELAYF",
                p_DEL_MODE  = "SCLK_ALIGNED",
                p_DEL_VALUE = "DELAY{}".format(rx_delay_taps),
                i_LOADN     = 1,
                i_MOVE      = 0,
                i_DIRECTION = 0,
                i_A         = pads.rx_ctl,
                o_Z         = rx_ctl_delayf),
]
```

This results in a design where all the IO_DELAY blocks are set to the same value. 
(The value set is actually 50 rather than 0 because of the `SCLK_ALIGNED` mode see nextpnr code [here](https://github.com/YosysHQ/nextpnr/blob/576baa994f4b9e1746a231469bfd9e65d1d5a2b4/ecp5/pack.cc#L1981) )

This creates a design that sometimes runs, but sometimes does not run because the timing is very marginal.

gregdavill:
For correct io delays in nextpnr the DEL_VALUE parameter needs to
be an integer, instead of the "DELAY{}" string.

The use of a "DELAY{}" string appears in the Lattice primitive
manual, but appears to be incorrect. At least based of the current
nextpnr.

Because we are not making use of dynamic io delays here we can
also use the simpler DELAYG block instead of DELAYF.

Fixes #50

### DIFF
--- a/liteeth/phy/ecp5rgmii.py
+++ b/liteeth/phy/ecp5rgmii.py
@@ -31,12 +31,9 @@ class LiteEthPHYRGMIITX(Module):
                 i1  = sink.valid,
                 i2  = sink.valid,
                 o   = tx_ctl_oddrx1f),
-            Instance("DELAYF",
+            Instance("DELAYG",
                 p_DEL_MODE  = "SCLK_ALIGNED",
-                p_DEL_VALUE = "DELAY0",
-                i_LOADN     = 1,
-                i_MOVE      = 0,
-                i_DIRECTION = 0,
+                p_DEL_VALUE = 0,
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Good catch, thanks!

PR #54, opened 2020-11-07
Fix syntax and load error in Wireshark etherbone dissector

rprinz08:
This patch fixes syntax errors for display of 'bytes' protocol fields and a loading error for the required 'bit' Lua module by modifying the module search path


enjoy-digital:
Thanks @rprinz08!

PR #56, opened 2020-12-22
Using multiple RMII phys with (possible multiple) external clock(s)

rprinz08:
- Control (disable) output of eth clock when using external clock
- Specify ClockDomain name when using multiple RMIIs each with its own external clock(domain)

### DIFF
--- a/liteeth/phy/rmii.py
+++ b/liteeth/phy/rmii.py
@@ -93,7 +93,8 @@ class LiteEthPHYRMIIRX(Module):
 
 
 class LiteEthPHYRMIICRG(Module, AutoCSR):
-    def __init__(self, clock_pads, pads, with_hw_init_reset):
+    def __init__(self, clock_pads, pads, with_hw_init_reset,
+                 no_clk_out=False):
         self._reset = CSRStorage()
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks. We can avoid the `no_clk_out` parameter by checking if `ref_clk` is present, ex: `hasattr(clock_pads, "ref_clk")`. For `clock_cd` I'm not sure to understand why both PHYs can't use the same `eth` reference clock?

rprinz08:
Agree, `hasattr(clock_pads, "ref_clk")` is nicer than `no_clk_out`. Multiple PHY clocks because they are located on modules which are used in the design, each with its own oscillator and the modules should not be modified ...

enjoy-digital:
@rprinz08: with https://github.com/enjoy-digital/liteeth/commit/dea35908c94873460643b48b162981590979e5e3 it's now possible to specify the reference clock domain through `refclk_cd` parameter. It's also possible to set `clock_pads=None` if you don't have `clock_pads` on your PHY.

rprinz08:
Thanks, problem solved.

PR #57, opened 2020-12-27
RMII PHY: Add support for ref_clk being provided externally

mwick83:
RMII spec allows an alternate mode in which ref_clk is an input on both
the MAC and the PHY. This is used by a no-name "LAN8720 ETH board"
break-out board. This commit allows this mode by specifying a ref_clk_in
Subsignal in clock_pads (instead of ref_clk).

### DIFF
--- a/liteeth/phy/rmii.py
+++ b/liteeth/phy/rmii.py
@@ -101,9 +101,16 @@ class LiteEthPHYRMIICRG(Module, AutoCSR):
         # RX/TX clocks
         self.clock_domains.cd_eth_rx = ClockDomain()
         self.clock_domains.cd_eth_tx = ClockDomain()
-        self.comb += self.cd_eth_rx.clk.eq(ClockSignal(refclk_cd))
-        self.comb += self.cd_eth_tx.clk.eq(ClockSignal(refclk_cd))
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @mwick83 for reporting the use case and initial implementation. I implemented things a bit differently in https://github.com/enjoy-digital/liteeth/commit/7448170390ddf4b7c353ae10932377255581c25a can you check it works for you? (You just have to provide `ref_clk` in `clock_pads` and set `refclk_cd` to `None` to use `clock_pads.ref_clk` as the RMII input reference clock).

mwick83:
Yes, that solution also works. I would have preferred my approach, because it made the mode of operation clear, just by defining the clock pads properly. With your soultion, I need to keep track of wether the ref_clk is an input or output during SoC creation seperate from the pure clock pads definition.

Anyway, thanks for the quick upstreaming of the use case!

enjoy-digital:
Thanks for the feedback @mwick83. I indeed saw the advantages of your approach but tried to reuse the parameters that were already in place and tried. This way refclk_cd parameter always has a meaning which was no longer the case with your solution.

PR #59, opened 2021-01-28
MAC: implement TX return channel & hardware timestamping

lschuermann:
This adds optional packet hardware timestamping to the SRAM-backed MAC interface. Hardware timestamping is a feature which can help with implementing standards such as IEEE 1588 Precision Time Protocol and can be interesting for a variety of time-critical domains. Compared to software-based timestamping of received and transmitted packets, having a hardware-timestamp available that corresponds to a specific point time in packet transmission & reception can greatly increase the accuracy of such protocols.

## Implementation strategy

While RX timestamping was very straightforward to implement (simply latch the current timestamp on reception start and store it in the SyncFIFO for incoming packet metadata), TX timestamping requires some more work: the MAC's user must be informed about the precise transmission timestamp after the packet has been sent. Up to now the MAC does not seem to have any return channel except for a pulse-triggered event indicating a packet has been transmitted.

Therefore, this PR also introduces a return-channel FIFO which contains metadata about transmitted packets. It changes the event source from pulse- to FIFO-ready-triggered. This seems to have the advantage of avoiding potentially missing a _TX done_ events if the CPU cannot handle & clear the first _TX done_ event prior to a second arrives, as the FIFO can queue `num_slots` _TX done_ events (sufficient for the maximum number of queueable packets).

The timestamp fields in the FIFO & CSRs for both RX & TX are generated only when a timestamp source signal is provided.

## Implications

Switching from a pulse-triggered event to a FIFO-based level-triggered event does of course change the behavior substantially. However, it seems that coincidentally some implementations (the bootloader and Tock's driver for LiteEth) continue to work just fine given that the event clearing semantics remain the same. I can imagine this causing issues where drivers we relying on the fact that events could not be queued (two packets will only require clearing _TX done_ once).

## Further Tasks / Questions

I'm no expert in Python and Migen so I appreciate input about coding style & other comments. It seems to work fine on an Arty and Sim, both with and without hardware timestamping, but probably requires careful review that I don't break anything.

I'm not sure whether this change is substantial enough for adding myself to any AUTHORS record.

Thanks!

### DIFF
--- a/liteeth/mac/__init__.py
+++ b/liteeth/mac/__init__.py
@@ -18,7 +18,8 @@ class LiteEthMAC(Module, AutoCSR):
         with_preamble_crc = True,
         nrxslots          = 2,
         ntxslots          = 2,
-        hw_mac            = None):
+        hw_mac            = None,
+        timestamp_source  = None):
         assert interface in ["crossbar", "wishbone", "hybrid"]
         self.submodules.core = LiteEthMACCore(phy, dw, endianness, with_preamble_crc)
         self.csrs = []
@@ -37,7 +38,13 @@ class LiteEthMAC(Module, AutoCSR):
             self.rx_slots  = CSRConstant(nrxslots)
             self.tx_slots  = CSRConstant(ntxslots)
             self.slot_size = CSRConstant(2**bits_for(eth_mtu))
-            self.submodules.interface = FullMemoryWE()(LiteEthMACWishboneInterface(32, nrxslots, ntxslots, endianness))
+            self.submodules.interface = FullMemoryWE()(LiteEthMACWishboneInterface(
+                dw = 32,
+                nrxslots = nrxslots,
+                ntxslots = ntxslots,
+                endianness = endianness,
+                timestamp_source = timestamp_source,
+            ))
             self.ev, self.bus = self.interface.sram.ev, self.interface.bus
             self.csrs = self.interface.get_csrs() + self.core.get_csrs()
             if interface == "hybrid":
--- a/liteeth/mac/sram.py
+++ b/liteeth/mac/sram.py
@@ -172,12 +172,16 @@ class LiteEthMACSRAMWriter(Module, AutoCSR):
 # MAC SRAM Reader ----------------------------------------------------------------------------------
 
 class LiteEthMACSRAMReader(Module, AutoCSR):
-    def __init__(self, dw, depth, nslots=2, endianness="big"):
+    def __init__(self, dw, depth, nslots=2, endianness="big", timestamp_source=None):
         self.source = source = stream.Endpoint(eth_phy_description(dw))
 
         slotbits        = max(log2_int(nslots), 1)
         lengthbits      = bits_for(depth*4)  # length in bytes
         self.lengthbits = lengthbits
+
+        # TX packet timestamping, for applications such as IEEE 1588
+        # Precision Time Protocol
+        timestampbits   = 0 if timestamp_source is None else value_bits_sign(timestamp_source)[0]
 
         # MAC Reader command channel
         self._start  = CSR()
@@ -188,6 +192,9 @@ class LiteEthMACSRAMReader(Module, AutoCSR):
 
         # MAC Reader return channel
         self._res_slot = CSRStatus(slotbits)
+        if timestamp_source is not None:
+            self._res_tx_timestamp = CSRStatus(timestampbits)
+
         self.submodules.ev = EventManager()
         self.ev.done       = EventSourceLevel()
         self.ev.finalize()
@@ -206,18 +213,28 @@ class LiteEthMACSRAMReader(Module, AutoCSR):
         ]
 
         # Result FIFO (return channel)
-        res_fifo = stream.SyncFIFO([("slot", slotbits)], nslots)
+        res_fifo_layout = [("slot", slotbits)]
+        if timestamp_source is not None:
+            res_fifo_layout += [("tx_timestamp", timestampbits)]
 
+        res_fifo = stream.SyncFIFO(res_fifo_layout, nslots)
         self.submodules += res_fifo
+
         self.comb += [
             res_fifo.source.ready.eq(self.ev.done.clear),
             self.ev.done.trigger.eq(res_fifo.source.valid),
             self._res_slot.status.eq(res_fifo.source.slot),
         ]
+        if timestamp_source is not None:
+            self.comb += self._res_tx_timestamp.status.eq(res_fifo.source.tx_timestamp),
 
         # Length computation
         read_address = Signal(lengthbits)
         counter      = Signal(lengthbits)
+
+        # Store the timestamp of transmission start
+        if timestamp_source is not None:
+            tx_start_timestamp = Signal(timestampbits)
 
         # FSM
         self.submodules.fsm = fsm = FSM(reset_state="IDLE")
@@ -225,9 +242,12 @@ class LiteEthMACSRAMReader(Module, AutoCSR):
             NextValue(counter, 0),
             If(cmd_fifo.source.valid,
                 read_address.eq(0),
+                *([] if timestamp_source is None
+                  else [NextValue(tx_start_timestamp, timestamp_source)]),
                 NextState("SEND")
             )
         )
+
         length_lsb = cmd_fifo.source.length[0:2]
         if endianness == "big":
             self.comb += If(source.last,
@@ -263,9 +283,9 @@ class LiteEthMACSRAMReader(Module, AutoCSR):
             NextState("IDLE")
         )
 
-        self.comb += [
-            res_fifo.sink.slot.eq(cmd_fifo.source.slot)
-        ]
+        self.comb += res_fifo.sink.slot.eq(cmd_fifo.source.slot)
+        if timestamp_source is not None:
+            self.comb += res_fifo.sink.tx_timestamp.eq(tx_start_timestamp)
 
         # Memory
         rd_slot = cmd_fifo.source.slot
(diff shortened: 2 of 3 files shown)


lschuermann:
It should be noted that when using a driver which does not necessarily acknowledge every packet's _done_ event, this might cause issues when switching to another system (as in the case of the bootloader), as the other system will receive the still pending _TX done_ events as soon as they are enabled (and won't know what to do with them since it hasn't sent any packets yet). A workaround is to acknowledge all pending events prior to enabling the event source again, but a packet might still be in transmission and generate an event accordingly.

I believe a proper fix would be a reset logic for both the SRAM reader and writer, aborting all current transactions and clearing the FIFOs. I'm not sure whether this is necessary though, simply requiring software to deal with unknown events (from the previous driver) might also be a valid solution.

enjoy-digital:
Thanks @lschuermann, this is a very nice contribution, the detailed description provided with the PR is really appreciated, makes things really clear and will simplify the review. I'm planning to merge other features in LiteEth next week so will have a closer look at your PR while  doing this work.

lschuermann:
Hi! @enjoy-digital! Did you get a chance to look into this?

liteeth/mac/__init__.py
```diff
@@ -37,7 +38,13 @@ def __init__(self, phy, dw,
...
             self.tx_slots  = CSRConstant(ntxslots)
             self.slot_size = CSRConstant(2**bits_for(eth_mtu))
-            self.submodules.interface = LiteEthMACWishboneInterface(32, nrxslots, ntxslots, endianness)
+            self.submodules.interface = LiteEthMACWishboneInterface(
+                32,
+                nrxslots,
+                ntxslots,
+                endianness,
+                timestamp_source
+            )
```
enjoy-digital: If we switch to multi-lines, can we specify the parameters: `dw=32`, `nrxslots= nrxslots`?

liteeth/mac/sram.py
```diff
@@ -62,23 +68,58 @@ def __init__(self, dw, depth, nslots=2, endianness="big"):
...
-        fifo = stream.SyncFIFO([("slot", slotbits), ("length", lengthbits)], nslots)
+        if timestamp_source is not None:
+            fifo = stream.SyncFIFO([
+                ("slot", slotbits),
+                ("length", lengthbits),
+                ("rx_timestamp", timestampbits),
+            ], nslots)
+        else:
+            fifo = stream.SyncFIFO([("slot", slotbits), ("length", lengthbits)], nslots)
         self.submodules += fifo
```
enjoy-digital, line 76: We can simplify using something like:

```python3
fifo_layout = [("slot", slotbits), ("length", lengthbits)]
if timestamp_source is not None:
    fifo_layout = [("rx_timestamp", timestampbits]
fifo = stream.SyncFIFO(fifo_layout, nslots)
```

liteeth/mac/sram.py
```diff
@@ -62,23 +68,58 @@ def __init__(self, dw, depth, nslots=2, endianness="big"):
...
+        # signal to hold the timestamp from the start of packet
+        # reception
+        if timestamp_source is not None:
+            rx_start_timestamp = Signal(timestampbits)
+
+            # Effectively build a latch, updating the timestamp only
+            # when starting the reception
+            self.comb += [
+                rx_start_timestamp.eq(rx_start_timestamp)
+            ]
```
enjoy-digital: You can use single line` self.comb +=  rx_start_timestamp.eq(rx_start_timestamp)` for this.

liteeth/mac/sram.py
```diff
@@ -62,23 +68,58 @@ def __init__(self, dw, depth, nslots=2, endianness="big"):
...
+                        ongoing.eq(1),
+                        NextValue(counter, counter + inc),
+                        NextState("WRITE")
+                    ).Else(
+                        NextValue(self.errors.status, self.errors.status + 1),
+                        NextState("DISCARD_REMAINING")
+                    )
+                )
+            )
+
```
enjoy-digital: We should avoid duplicating too much FSM code, this can be simplified with something like:
````python3          
    fsm.act("IDLE",
                If(sink.valid,
                    If(fifo.sink.ready,
                        ongoing.eq(1),
                        *[] if timestamp_source is None else *[rx_start_timestamp.eq(timestamp_source)],
                        NextValue(counter, counter + inc),
                        NextState("WRITE")
                    ).Else(
                        NextValue(self.errors.status, self.errors.status + 1),
                        NextState("DISCARD_REMAINING")
                    )
                )
            )
````

File: liteeth/mac/sram.py [hunk 5]
```diff
@@ -145,50 +195,99 @@ def __init__(self, dw, depth, nslots=2, endianness="big"):
...
+            self._level.status.eq(cmd_fifo.level)
+        ]
+
+        # Result FIFO (return channel)
+        if timestamp_source is not None:
+            res_fifo = stream.SyncFIFO([("slot", slotbits), ("tx_timestamp", timestampbits)], nslots)
+        else:
+            res_fifo = stream.SyncFIFO([("slot", slotbits)], nslots)
+
+        self.submodules += res_fifo
```
enjoy-digital, line 221: Same remark than for the rx fifo.

liteeth/mac/sram.py
```diff
@@ -145,50 +195,99 @@ def __init__(self, dw, depth, nslots=2, endianness="big"):
...
-            self._ready.status.eq(fifo.sink.ready),
-            self._level.status.eq(fifo.level)
+            res_fifo.source.ready.eq(self.ev.done.clear),
+            self.ev.done.trigger.eq(res_fifo.source.valid),
+            self._res_slot.status.eq(res_fifo.source.slot),
         ]
+        if timestamp_source is not None:
+            self.comb += [
+                self._res_tx_timestamp.status.eq(res_fifo.source.tx_timestamp),
+            ]
```
enjoy-digital, line 230: Avoid multi-lines here.

liteeth/mac/sram.py
```diff
@@ -145,50 +195,99 @@ def __init__(self, dw, depth, nslots=2, endianness="big"):
...
 
+        # Store the timestamp of transmission start
+        if timestamp_source is not None:
+            tx_start_timestamp = Signal(timestampbits)
+
+            # Effectively build a latch, updating the timestamp only
+            # when starting the transmission
+            self.comb += [
+                tx_start_timestamp.eq(tx_start_timestamp)
+            ]
```
enjoy-digital: Avoid multi-lines here.

liteeth/mac/sram.py
```diff
@@ -145,50 +195,99 @@ def __init__(self, dw, depth, nslots=2, endianness="big"):
...
+        else:
+            fsm.act("IDLE",
+                NextValue(counter, 0),
+                If(cmd_fifo.source.valid,
+                    read_address.eq(0),
+                    NextState("SEND")
+                )
+            )
+
+        length_lsb = cmd_fifo.source.length[0:2]
```
enjoy-digital, line 251: Avoid duplicating FSM code here with similar changes than in the other FSM.

enjoy-digital (changes requested):
Hi @lschuermann, sorry for the delay, I wanted to review it sooner and get back to you. I did a first review listing the few things I see that would minimize changes in the code for this PR. This is mostly coding style remarks that I try to apply to have minimal code duplication and ease maintenance/evolution. Would you mind doing these changes? (I could also do it if you  prefer). I'll have another closer review at the PR later, but I see no reasons we wouldn't be able to merge it soon.

liteeth/mac/sram.py
enjoy-digital: In fact, we should avoid latches, so you can probably remove this code and use `NextValue(rx_start_timestamp, timestamp_source)` in the FSM.

liteeth/mac/sram.py
enjoy-digital: In fact, we should avoid latches, so you can probably remove this code and use NextValue(tx_start_timestamp, timestamp_source) in the FSM.

enjoy-digital:
I did a closer review and think the implementation is OK, we would just need to fix the few implementation details. Please let me know if you want help on this.

liteeth/mac/sram.py
lschuermann: Ah, this is cool. I'm still very new to Migen and haven't used Python a lot, so I wanted to "just make it work". :)

lschuermann:
Thanks for the review. I addressed the requested changes. I'm not exactly a Python expert so it's great to get to know more of the syntactic sugar and quirks it has.

Yet to test these changes, will do later today.

enjoy-digital:
Thanks @lschuermann, once you confirm the updates version works, I could merge it and do the potential other minor changes myself.

lschuermann:
@enjoy-digital Terribly sorry to have kept you waiting this long. Unfortunately, I'm currently not at my usual location and ran into some unexpected issues with my hardware setup.

I now can, however, confirm that these changes work in the simulation and on the hardware. Given I'm currently researching in this area, there _might_ be some improvements later, but the interface for timestamps themselves seems to work fine. I'd be really happy if they can make it into LiteEth upstream. :)

enjoy-digital:
@lschuermann: No problem, I also often need time to review things. I'll review and merge your update PR in the next days.

lschuermann:
Rebased these changes as there was a merge conflict:

```
<<<<<<< upstream/master
            self.submodules.interface = FullMemoryWE()(LiteEthMACWishboneInterface(32, nrxslots, ntxslots, endianness))
=======
            self.submodules.interface = LiteEthMACWishboneInterface(
                dw = 32,
                nrxslots = nrxslots,
                ntxslots = ntxslots,
                endianness = endianness,
                timestamp_source = timestamp_source,
            )
>>>>>>> 534699e
```

I don't know what wrapping `LiteEthMACWishboneInterface` in `FullMemoryWE()` does as I found absolutely no documentation on this, hence I've just mimicked this with my changes.

enjoy-digital (approved):
Thanks @lschuermann for the update and sorry for the delay. This looks good! I'm going to merge and do minor changes.

enjoy-digital:
I reworked the code a bit with https://github.com/enjoy-digital/liteeth/commit/392414eef8991ddea619a0a21d5d2ec7adf3c9ae, can you check Timstamping is still working fine with your application? For the DMA Reader, I have a preference to disable the Status channel when Timestamping is not used to avoid impacting the CSR Mapping (which would cause issues with the different software drivers using LiteEth).

Otherwise that would indeed be good to add you in the list of authors of sram.py, can you provide me the name/mail you want to use in the copyright?

Thanks.

PR #61, opened 2021-02-16
Fix "is not" literal SyntaxWarning

shenki:
Signed-off-by: Joel Stanley <joel@jms.id.au>


enjoy-digital:
Thanks @shenki! This is merged.

PR #64, opened 2021-03-11
allow for different nrxslots and ntxslots

mczerski:
Background:
When there is a lot of broadcasts in the network, receive buffers
may overflow easly. Especially having onlu 2 of them.
To prevent that you can enlarge nrxslots, but because
nrxslots + ntxslots must be the power of two, you must also
inrease ntxslots. But there is no need to have more than 2 tx
buffers (they work as ping-pong buffer), the CPU will not use
more than two buffers.

So being able to set for example nrxslots=8 and ntxslots=2
is quite reasonable.


enjoy-digital (approved):
Thanks @mczerski, this looks good.

PR #65, opened 2021-03-29
Revert "phy/ecp5rgmii: remove p_DEL_MODE (not required since we speci…

shuffle2:
This was silently (😞) causing my design to stop functioning correctly* (using Diamond 3.11). The lattice tools _will_ halt with error if you try to use the Reveal Inserter with `USER_DEFINED` and the `DEL_VALUE`s liteeth is setting.

Just reverting this commit seems to be the easiest way to fix behavior and prevent tools from erroring, although tbh I haven't looked further into the problem.

image

*Ethernet traffic was either not sent on the wire from the phy, or whatever was sent was not understood by my host PC

### DIFF
--- a/liteeth/phy/ecp5rgmii.py
+++ b/liteeth/phy/ecp5rgmii.py
@@ -32,6 +32,7 @@ class LiteEthPHYRGMIITX(Module):
                 i2  = sink.valid,
                 o   = tx_ctl_oddrx1f),
             Instance("DELAYG",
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @shuffle2, nice catch. Yosys/NextPnr was used for the last tests and probably ignores `DEL_MODE` when `DEL_VALUE` is set.

PR #67, opened 2021-04-27
phy/s7rgmii: add configurable iodelay_clk_freq

jedrzejboczar:
This allows to avoid a conflict when we want to use different iodelay frequency then 200 MHz for DRAM.


enjoy-digital:
Thanks @jedrzejboczar, this is merged. BTW, I don't know if you are aware of it, but possible to have multiple IDELAYCTRLs (one per bank) but we have to define the groups and pass an attribute to the I/ODELAY instance, which is not very convenient, so with our SoCs your solution is probably easier.

PR #73, opened 2021-08-10
liteeth/phy: add configurable hw reset duration

jedrzejboczar:
Sometimes 256 cycles is to short to reset the Ethernet PHY. This PR makes the reset duration configurable, leaving the previous value as the default.


enjoy-digital:
Thanks @jedrzejboczar! This is merged.

PR #72, opened 2021-08-08
mac: Allow configuring usage of FullMemoryWE (fixes #70)

david-sawatzke:
On ecp5 `FullMemoryWE` leads to an increase of DP16KD block mem, while
it works better on Intel/Altera devices according to
6c3af746e28f55089c9f79966205499af394ae34.

Simple solution: Make it configurable

EDIT: Original issue leading to `FullMemoryWE` being used is #63

PR #49, opened 2020-10-11
mac/core: Improve timing closure of core

gregdavill:
On ECP5 targets the core struggles to meet timing closure. This
change adds buffers to the CRC module on tx/rx paths.
This results in 20-30MHz gain to max clock rate.

This fixes #47

This is the timing result I've just got from running:
```console
$ ./versa_ecp5.py --build --with-ethernet --integrated-rom-size 36864 
...
Info: Max frequency for clock                     '$glbnet$sys_clk': 89.22 MHz (PASS at 75.01 MHz)
Info: Max frequency for clock '$glbnet$eth_clocks_rx$TRELLIS_IO_IN': 148.57 MHz (PASS at 125.00 MHz)
Info: Max frequency for clock                    '$glbnet$init_clk': 314.37 MHz (PASS at 25.00 MHz)
Info: Max frequency for clock        '$glbnet$clk100$TRELLIS_IO_IN': 451.06 MHz (PASS at 100.00 MHz)
...
```

PR #21, opened 2019-11-13
Add hooks and features for 10/25G implementation

jersey99:
Liteeth is very feature rich. However, lacking >1G functionality. Following modifications have made a simple working 10G liteth core that can be tested on hardware.

For [motivation](https://docs.google.com/document/d/1toVDIePKeY9EhktqZA0PLd1TKcjGvYjkx7VV2SNyDA8/edit?usp=sharing)

Liteeth lacked a MAC implementation of datawidth > 8bits.  So, in order to facilitate wider lanes for XGMII, extended the datawidths of all the modules (PHY, MAC, ARP, IP, and all their supporting modules etc have been modified.

1. The [XGMII BASE-R chapter 46](http://www.ismlab.usf.edu/dcom/Ch3_802.3-2005_section4.pdf#G3.31950) phy implementing the incomplete XGMII standard is in liteeth/phy/xgmii.py
2. Liteth written in Migen, can be converted to Verilog which is simulated using Verilator. The Linux TUN/TAP user space interface mixed with Verilator simulation written in C/C++ takes a stream of data from LiteEthMAC/PHY, and is sent to the Linux Kernel's virtual networking interface (and vice versa).  Since the Linux Kernel's TAP interface only provides MAC and higher layers, the PHY interface can actually be tested in the C/C++ layer by decoing PHY signals into data words before passing it into the TAP interface. This is tested in simulation over verilator and tapcfg at litex/build/sim/core/modules/xgmii_ethernet/xgmii_ethernet.c This is an extension of the existing ethernet.c test. The xgmii_ethernet.c can perhaps be extended to all widths and controls for MII/GMII/RGMII and perhaps even the PCS/PMA.
3. Liteeth uses streaming interface as interconnect between modules. When modules have byte wide boundaries and the datawidth remains 1byte, the interconnect is arguably “simpler”. When the datawidth is higher, the header-payload boundaries don't align with datawidth boundaries, which requires modifications to the packetizers and streams.  Now implemented at XGMII branch on [litex](https://github.com/jersey99/litex/tree/xgmii) and [liteeth](https://github.com/jersey99/liteeth/tree/xgmii)
Both 1G and 10G use the same codepath.
[...]

PR #74, opened 2021-08-10
Add toggleable 32 bit support to mac pipeline

david-sawatzke:
This PR adds 32 bit support to all components of the mac pipeline. For everything apart from the crc 64 bit should also work but is *currently* untested. 64 bit support for the crc shouldn't be hard to add without a completely separate codepath, but since I wanted to keep this PR smaller and can't test it right now, I didn't implement it.

This PR duplicates some things from #21, but does some things nicer IMHO, like the preamble. #21 also doesn't properly implement crc for the 32 bit and padding doesn't work. It's also broken on big endian systems. I've also tried to add a few comments when they're helpful.

#21 also seems to have a trick to reduce the amount of crc engines needed, but this looked very confusing to me, so I took the simple route.

Each commit should be in a working state on little endian systems, it's probably also easier to review it commit-by-commit.

I'm not sure if there's a better way to do the endian conversion, I haven't been able to find one. We may want to do the endian conversion in any case for 32/64 bit phy support and make the stride converter only support little endian.

With this PR in place, the colorlight target easily hits the 125MHz cd target (this is with #72):
```
Info: Max frequency for clock '$glbnet$eth_clocks0_rx$TRELLIS_IO_IN': 138.29 MHz (PASS at 125.00 MHz)
```

With a smaller depth for the stride converter, it gets even higher. I'm not sure why the stride converter has such a large buffer, especially tx. With stride converter depth set at 4, the ethernet core also hits the frequency constraints on my own design all the time with > 150 MHz)

#49 can probably also be reverted.

### DIFF
--- a/liteeth/mac/__init__.py
+++ b/liteeth/mac/__init__.py
@@ -20,9 +20,10 @@ class LiteEthMAC(Module, AutoCSR):
         ntxslots          = 2,
         hw_mac            = None,
         timestamp         = None,
-        full_memory_we    = False):
+        full_memory_we    = False,
+        sys_data_path     = True):
         assert interface in ["crossbar", "wishbone", "hybrid"]
-        self.submodules.core = LiteEthMACCore(phy, dw, endianness, with_preamble_crc)
+        self.submodules.core = LiteEthMACCore(phy, dw, endianness, with_preamble_crc, sys_data_path)
         self.csrs = []
         if interface == "crossbar":
             self.submodules.crossbar     = LiteEthMACCrossbar(dw)
--- a/liteeth/mac/core.py
+++ b/liteeth/mac/core.py
@@ -61,6 +61,11 @@ class LiteEthMACCore(Module, AutoCSR):
             self.sync += If(self.ps_preamble_error.o,
                             self.preamble_errors.status.eq(self.preamble_errors.status + 1)),
 
+        if sys_data_path:
+            self.data_path_converter(tx_pipeline, rx_pipeline, core_dw, phy.dw, endianness)
+            cd_tx = cd_rx = "sys"
+            dw = core_dw
+
         if not isinstance(phy, LiteEthPHYModel) and with_preamble_crc:
             # CRC insert/check
             crc32_inserter = BufferizeEndpoints({"sink": DIR_SINK})(crc.LiteEthMACCRC32Inserter(eth_phy_description(dw)))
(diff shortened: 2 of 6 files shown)


david-sawatzke:
Additional notes for testing with colorlight target:
32 bit data path (with #72):
```
Info: Device utilisation:
Info: 	       TRELLIS_SLICE:  6275/12144    51%
Warning: Max frequency for clock             '$glbnet$main_crg_clkout0': 59.35 MHz (FAIL at 60.00 MHz)
Info: Max frequency for clock '$glbnet$eth_clocks0_rx$TRELLIS_IO_IN': 146.01 MHz (PASS at 125.00 MHz)
```

Regression testing for 8 bit data path (#72):
```
Info: Device utilisation:
Info: 	       TRELLIS_SLICE:  5417/12144    44%
Info: Max frequency for clock             '$glbnet$main_crg_clkout0': 60.88 MHz (PASS at 60.00 MHz)
Info: Max frequency for clock '$glbnet$eth_clocks0_rx$TRELLIS_IO_IN': 126.31 MHz (PASS at 125.00 MHz)
```
On master (with #72):
```
Info: Device utilisation:
Info: 	       TRELLIS_SLICE:  5348/12144    44%
Warning: Max frequency for clock             '$glbnet$main_crg_clkout0': 51.12 MHz (FAIL at 60.00 MHz)
Info: Max frequency for clock '$glbnet$eth_clocks0_rx$TRELLIS_IO_IN': 131.03 MHz (PASS at 125.00 MHz)
```

Note: The timings vary a lot, I've previously had a lot of issues hitting the eth clock target on master. In all cases, the cpu was the critical path for sys_clk. Utilization for everything apart from `TRELLIS_SLICE`s stayed the same.

enjoy-digital:
Thanks @david-sawatzke! This looks very good. I'm not able to review it carefully now (vacation) but we'll be able before the end of August.

Regarding the depth of the Converter, I also think this can be reduced now. In the past, the SRAM Reader was not efficient 
but this has been fixed with https://github.com/enjoy-digital/liteeth/commit/081bf46ca6f17721ffe538d2c19707c44a874045 and I suspect this was the reason TX depth had to be set to 64. With the fix we should now be able to reduce TX depth to the minimum.

david-sawatzke:
> With the fix we should now be able to reduce TX depth to the minimum.

Sounds good! I'll leave it out of this PR, it's already large enough as-is

david-sawatzke:
I looked a bit at testing and it does already get tested! (more-or-less by accident) The 8 bit mac path is tested via `test_mac_core.py` and the 32 bit mac path gets tested via `test_mac_wishbone.py`. Misusing them accidentally in this way is not great, but at least works. I also looked into adapting `test_mac_core.py` to also test a 32 bit path, but `PacketLogger` is very much not set up for supporting `last_be`  and doing so feels hairy, with all the points where it gets used.

enjoy-digital:
Sorry for the delay. The changes look good in general so we can merge. I'll do some tests with the other PRs and will maybe do minor changes. Thanks a lot @david-sawatzke for this work!

PR #75, opened 2021-08-14
mac/{sram,wishbone}: remove assumptions about 32-bit dw and add SRAM to Wishbone dw conversion

lschuermann:
This PR is introducing two commits as a foundation of integrating a 64-bit MAC data path:

In `liteeth/mac/sram.py`, various assumptions about having a 32 bit SRAM data width are removed. Especially the last_be encoding and decoding have been separated into a module, which generates the one-hot encodings and decodings on the fly. Furthermore, the TX counter increments and SRAM port address width are now dependent on the data width of the SRAM.

Furthermore, to actually make a 64-bit SRAM data width usable on 32-bit SoCs, `liteeth/mac/wishbone.py` now integrates a Wishbone bus data width conversion. If the SRAM data width matches the Wishbone bus data width, this should not produce any additional overhead as this case is handled in `wishbone.Converter`. Otherwise, an up- or down-conversion module is instantiated accordingly. This results in the entire SRAM port always being fully exposed in the SoC's expected data width.

These changes are submitted in an individual PR for easy review.

### DIFF
--- a/liteeth/mac/sram.py
+++ b/liteeth/mac/sram.py
@@ -1,10 +1,13 @@
 #
 # This file is part of LiteEth.
 #
+# Copyright (c) 2021 Leon Schuermann <leon@is.currently.online>
 # Copyright (c) 2015-2020 Florent Kermarrec <florent@enjoy-digital.fr>
 # Copyright (c) 2015-2018 Sebastien Bourdeauducq <sb@m-labs.hk>
 # Copyright (c) 2017 whitequark <whitequark@whitequark.org>
 # SPDX-License-Identifier: BSD-2-Clause
+
+from math import log2, ceil
 
 from liteeth.common import *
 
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @lschuermann, I started having a closer look at the different LiteEth PRs. For this PR, regarding the SRAM implementation it seems fine, but I'm wondering about the use-case: The Soft Core CPU are already not able to saturate the 1Gbps link so I'm wondering what are your plans with 10Gbps Ethernet and Soft Core CPU?

Otherwise, for the data-width adaptation, LiteX is already able to do it automatically, so we'll be able to avoid it here.

lschuermann:
> The Soft Core CPU are already not able to saturate the 1Gbps link so I'm wondering what are your plans with 10Gbps Ethernet and Soft Core CPU?

I'm aware the Soft Core CPU would not be able to catch up, but I still need to use it for management traffic. However, when the entire MAC pipeline is 64 bit, we need to have a 64 bit path into the SRAM such that we don't need to buffer the entire packet maximum length * receive slots before the SRAM. Same goes for TX, we need to send the packets in one go without interruption and thus we would need to buffer at least one full frame such that we can send all 64 bit bus words in a single transaction.

Hence it seems to make a lot of sense to just let the SRAM interface for the MAC be as large as the rest of the MAC pipeline.



> Otherwise, for the data-width adaptation, LiteX is already able to do it automatically, so we'll be able to avoid it here.

I'm afraid I don't quite get that. Can you elaborate on what you mean?

enjoy-digital:
My question was more related to the use case than the technical aspects (with which I totally agree).  One way to add complexity to a project is to support use cases that are not useful (or not yet), so I'm just trying to understand what's your use case at a system level. Maybe we can discuss this directly together.

Regarding the automatic data-width adaptation, it's already possible to add 64-bit peripherals to a 32-bit SoC with LiteX, the adapter will automatically be inserted:
https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc.py#L290-L343

david-sawatzke:
In my project I'm using gigabit data transfer, where large parts of the network activity are (read: will be) handled by hardware (because the softcore would never be able to keep up), but some management (a telnet console) and less bandwidth intensive tasks are handled by the softcore. 

For 10 gig you could also imagine having a large data pipe with the hardware stack, but some tcp transfers are handled by the softcore (because it's more flexible). Without a wider sram pipeline it's also impossible to boot over tftp with a 10 gig network interface.

You'll never need the bandwidth for the softcore, but otherwise it's not possible to have the hybrid interface.

lschuermann:
@david-sawatzke perfectly summarized what I've just discussed with @enjoy-digital via IRC. Indeed the use cases lie within usage of the Hybrid interfaces. The softcores will never be able to handle the bandwidth which this 64-bit interface would give us, but it's conceptually much simpler than to introduce buffering somewhere in between.

As discussed with @enjoy-digital, I'll push some fixes to this code today.

enjoy-digital:
Thanks @david-sawatzke, as discussed @lschuermann this makes sense for this use case. I'll merge the sram adaptation. The dw conversion should not be required.

enjoy-digital:
@lschuermann: I'm merging the full PR and will probably rework the dw conversion, Thanks!

enjoy-digital:
@lschuermann: I just did the dw conversion rework with https://github.com/enjoy-digital/liteeth/commit/cab4e93317cb3656153c5bedc9b412d733e9a7d2. This would still need to be tested on hardware for 64-bit but the data-with conversion will now automatically be done by LiteX integration code [here](https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc.py#L293-L308) and avoid duplication of such features in the codebase.

lschuermann:
@enjoy-digital Awesome, thank you. I'll have access to the board to test next week. I'll try to test in the simulator beforehand. :)

PR #76, opened 2021-08-14
Add 64-bit XGMII PHY implementation for 10G Ethernet

lschuermann:
Adds support for 64-bit wide XGMII PHYs in LiteEth. A 64-bit wide XGMII data path is a common method to interconnect multi-gigabit Ethernet inside FPGAs. This module expects a 64-bit MAC data path, which is to be added later. It has been tested locally using a rewritten XGMII module for the LiteX simulator as well as on a KCU116 board.

This work has been inspired by enjoy-digital/liteeth#21 but is entirely rewritten using Migen FSMs and with respect to IEEE802.3-2018. Thanks to Florent Kermarrec (@enjoy-digital) and Vamsi Vytla (@jersey99) for providing a basic implementation.

This implementation does not yet support proper 32-bit (DDR) XGMII PHYs, although support can be easily added by an additional module which performs the DDR encoding / decoding of the data respectively.

### DIFF
--- a/liteeth/phy/__init__.py
+++ b/liteeth/phy/__init__.py
@@ -26,6 +26,7 @@ from liteeth.phy.mii  import LiteEthPHYMII
 from liteeth.phy.rmii import LiteEthPHYRMII
 from liteeth.phy.gmii import LiteEthPHYGMII
 from liteeth.phy.gmii_mii import LiteEthPHYGMIIMII
+from liteeth.phy.xgmii import LiteEthPHYXGMII
 
 from liteeth.phy.s6rgmii import LiteEthPHYRGMII as LiteEthS6PHYRGMII
 from liteeth.phy.s7rgmii   import LiteEthPHYRGMII as LiteEthS7PHYRGMII
--- a/liteeth/phy/xgmii.py
+++ b/liteeth/phy/xgmii.py
@@ -1,0 +1,321 @@
+#
+# This file is part of LiteEth.
+#
+# Copyright (c) 2021 Leon Schuermann <leon@is.currently.online>
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+
+from migen import Module
+from liteeth.common import *
+
+from functools import reduce
+from operator import or_
+
+XGMII_IDLE = Constant(0x07, bits_sign=8)
+XGMII_START = Constant(0xFB, bits_sign=8)
+XGMII_END = Constant(0xFD, bits_sign=8)
+
+class LiteEthPHYXGMIITX(Module):
+    def __init__(self, pads, dw):
+        # Enforce 64-bit data path
+        assert dw == 64
+
+        # Sink for data to transmit
+        self.sink = sink = stream.Endpoint(eth_phy_description(dw))
+
+        # Transmit FSM
+        self.submodules.fsm = fsm = FSM(reset_state="IDLE")
+
+        fsm.act("IDLE",
+            If(sink.valid,
+                # Currently idling, but a new frame is ready for
+                # transmission. Thus transmit the preamble, but replace the
+                # first byte with the XGMII start of frame control
+                # character. Accept more data.
+                pads.tx_ctl.eq(0x01),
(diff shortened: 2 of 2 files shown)


lschuermann:
Converting to draft PR, as I have observed some irregular behavior when running on the KCU116 FPGA board.

lschuermann:
Alright, it seems to work reliably now. The issue wasn't the hardware, or different behavior in simulation compared to the hardware. Rather, I must've misread the specification of the XGMII interface as defined by IEEE802.3. _IEEE802.3-2018 Section Eight, 126.3.2.2.10 Start_ regulates that the XGMII start control character is only valid on the first octet of a given XGMII bus word. However, the standard does not relate to the 64-bit bus word (single data rate) as present in the FPGA, but rather the 32-bit bus word (double data rate) as used in physical XGMII interfaces. Thus, on a 64-bit interface the XGMII start control character is valid on both the first and the fifth octet of the bus word.

Because handling this (and the resulting offsets + buffering) in the RX implementation makes for very complex code, this implementation takes a different approach. It introduces an Aligner which makes sure the XGMII start of frame control character is always present on the first octet in a 64-bit bus word.

However, such an Aligner may require creating two distinct aligned frame contents from a single 64-bit XGMII bus word in a single clock cycle, like so:

```
64-bit XGMII BUS Word    -> Aligned Frame #0        | Aligned Frame #1
 0  0  0  1  1               0  0  0  1  1  1  1  1 |  1  0  0  0
XX XX XX FD FB YY YY YY     XX XX XX FD 07 07 07 07 | FB YY YY YY ... <NEXT_BUS_WORD[0:32]>
Frame-----> <-----Frame     Frame----->               <-----Frame
```

While the aligned frame No. 1 will be delayed by one clock cycle, once we're observing an XGMII start of frame control character at the first octet of a 64-bit bus word again we have two 64-bit XGMII words ready simultaneously -- which we can't handle.

However, this does not seem to be an issue, given Ethernet mandates an interpacket gap of 96 bits, which is sufficient to make a case such as the one above impossible to occur. While for technical reasons the interpacket gap at the receiver may be reduced below the interFrameSpacing mandated for the transmitter (IPG shrinkage, as explained in IEEE802.3-2018 Section One, 13.4.2 Interpacket gap (IPG) shrinkage), the interpacket gap at the receiver can not be lower than 5 bytes. Thus we can always properly align each and every packet to the start of a 64bit bus word, independent on the particular alignment of the previous packet, we should never need to discard valid data.

One may rewrite the RX PHY at some point to directly handle these cases and potentially have a more graceful scheme of handling these error conditions, however this seems to be a good approach for now and really reduces the implementation complexity of the PHY.

I may well have missed something, please feel free to comment. Thanks to @david-sawatzke for helping debug this.

jersey99:
@lschuermann Thanks a lot for all the work! I do like the FSM in there. I will read through it, and update my project for KC705 with your xgmii.py and look at some of the simulations you have.

I will wait for @enjoy-digital to comment on what you mean by 64 bit data path, I believe this part is tried and tested as of a few months ago.

lschuermann:
@jersey99 thanks for the feedback! Let me know whether you run into any issues, also with the simulation.

Regarding the 64bit XMGII bus word discussion: with my 10G Ethernet card (Aquantia) this problem only occurs for certain packet lengths and presumably is also highly dependant on the PHY used (or instantiated in the FPGA). For my development efforts I'm using a PHY of the verilog-ethernet project which connects to the GTY transceivers of my FPGA. Presumably some combinations of these variables will cause the XGMII control character to be sent on the fifth Byte of the 64bit XGMII word, which would be the first Byte of a 32bit XGMII DDR bus word. This is technically allowed by the standard.

This is also why this issue was so incredibly hard to debug. The problems manifested themselves only with certain packet lengths (as that presumably influences buffer handling & timing in my NIC) as well as certain NICs, etc. Some combinations will just always send the XGMII start of frame control character at the first XGMII bus word byte (i.e. rising edge of the clock for 32bit DDR), which your implementation did handle just fine.

enjoy-digital:
Thanks a lot @lschuermann, this is merged.

PR #78, opened 2021-09-01
phy/gmii: add model parameter to skip clock buffers & generation

lschuermann:
To support a simple GMII simulation, skip clock generation and buffer logic. This allows to operate a GMII interface over sys_clk. Proper GMII clocking support can still be added in the simulation, this should work when setting model = False.

It also sets an attribute "model" such that we can avoid adding Platform constraints in the rest of the ecosystem (such as `litex/litex/soc/integration/soc.py`, `add_ethernet` and `add_etherbone`).

### DIFF
--- a/liteeth/phy/gmii.py
+++ b/liteeth/phy/gmii.py
@@ -47,7 +47,7 @@ class LiteEthPHYGMIIRX(Module):
 
 
 class LiteEthPHYGMIICRG(Module, AutoCSR):
-    def __init__(self, clock_pads, pads, with_hw_init_reset, mii_mode=0):
+    def __init__(self, clock_pads, pads, with_hw_init_reset, mii_mode=0, model=False):
         self._reset = CSRStorage()
 
         # # #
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @lschuermann, the resulting PHY when configured as a model seems very similar to the `LiteEthPHYModel` (https://github.com/enjoy-digital/liteeth/blob/master/liteeth/phy/model.py) that is currently used in simulation, so I'm not sure to understand the advantage of your approach vs the current one, can you provide more information?

lschuermann:
This is to be used in conjuction with enjoy-digital/litex#1023. It really helped us to have a proper 8 bit PHY interface into the simulation. This is because, in constrast to the `LiteEthPHYModel`, it uses the entire pipeline as defined in the MAC, including for example CRC. The simulation contains code to generate and properly verify the CRC which can be a great tool to spot errors in the implementation.

In general, it seems good to at least have the option of a simulated PHY which is very close to actual HW, even if you wouldn't necessarily use apart from testing LiteEth.

enjoy-digital:
@lschuermann: Thanks, I can understand the use-case but I still don't understand the difference with `LiteEthModel` which was created with the same intent. The only difference I can see between `LiteEthModel`and `LiteEthPHYGMII` with `model=True` is that the pads's name are different:
- source_valid : tx_en
- source_data : tx_data
- sink_valid : rx_dv
- sink_data : rx_data

If you have a preference for a naming similar to the one used by the hardware, I would rather extend LiteEthPhyModel to also support it. Would it make sense for you?

lschuermann:
As far as I know the LiteEthModel does not include the Ethernet CRC or preamble. It transmits only raw packet data. 

The GMII simulation properly simulates an actual GMII interface where these things are present and checked, both in the simulation and in LiteEth.

lschuermann:
I can offer to look into making the GMII simulation work with proper RX and TX clocks which might make this unnecessary. Would that be a good option?

enjoy-digital:
When you do a simulation with LiteEthModel, the simulated hardware code is the same than with `LiteEthPHYGMI`I with `model=True`. (CRC and Preamble are also included in the simulation). That's the Verilator software stub that is then different.

Initially we were only doing simulations with LiteEthModel (which is equivalent to a GMII PHY). What I see is that we'll now have some duplication that we should probably avoid. Since we are going to simulate other PHYs (XGMII) it could make sense to remove the `LiteEthPHYModel` and just use the new model mode for PHYs supporting it.

I'll think about it, not need to spend time trying to get the GMII simulation with un-modified CRG for now.

lschuermann:
> When you do a simulation with LiteEthModel, the simulated hardware code is the same than with `LiteEthPHYGMI`I with `model=True`. (CRC and Preamble are also included in the simulation). That's the Verilator software stub that is then different.

I'm afraid I don't quite follow. To me this looks like the preamble and CRC parts of the MAC are not even generated:

https://github.com/enjoy-digital/liteeth/blob/1db5299a62638b66f9b887baac78f59573cde2ac/liteeth/mac/core.py#L32-L37

> Initially we were only doing simulations with LiteEthModel (which is equivalent to a GMII PHY). What I see is that we'll now have some duplication that we should probably avoid. Since we are going to simulate other PHYs (XGMII) it could make sense to remove the `LiteEthPHYModel` and just use the new model mode for PHYs supporting it.

I would appreciate that. It seems good to just commit to some standard and not use something custom, which is similar to a standard. Additionally, I'm of course biased on this one, I do think that in terms of the Verilator module, the GMII implementation is quite a bit nicer than what we had previously. We can maybe measure the overhead which the preamble and CRC hardware have compared to the model implementation. If it's insignificant, I'd be very happy to get rid of the LiteEthPHYModel in the long term.

enjoy-digital:
Thanks @lschuermann, as discussed together, I'm merging the PR and once all the different PR will be merge, we could discuss which models really have to be kept.

PR #80, opened 2021-09-10
Make LiteEthMACCore clock domain crossing TX/RX fifo depths configurable

rprinz08:
Commit 72dd7bf reduced buffering from 64 to 32 which affects a design. So make it configurable to be flexible.

### DIFF
--- a/liteeth/mac/core.py
+++ b/liteeth/mac/core.py
@@ -17,7 +17,8 @@ from litex.soc.interconnect.stream import BufferizeEndpoints, DIR_SOURCE, DIR_SI
 # MAC Core -----------------------------------------------------------------------------------------
 
 class LiteEthMACCore(Module, AutoCSR):
-    def __init__(self, phy, dw, endianness="big", with_preamble_crc=True, with_padding=True):
+    def __init__(self, phy, dw, endianness="big", with_preamble_crc=True, with_padding=True,
+                 cdc_tx_depth=32, cdc_rx_depth=32):
         if dw < phy.dw:
             raise ValueError("Core data width({}) must be larger than PHY data width({})".format(dw, phy.dw))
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @rprinz08, can you provide more information on the affected design? (especially the sys_clk_freq of the SoC and speed of the Ethernet Link)

rprinz08:
The design runs with sys_clk_freq=100MHz and uses a LiteEthPHYRMII with external 50MHz clock and a LiteEthMACCore. Interface runs at 100Mbit full duplex. With reduced CDC depth, no last signal is issued by LiteEthMACCore source when a network packet was completely received.

enjoy-digital:
@rprinz08 Thanks for the info. Rather than exposing the FIFO depths, I would like to understand the real cause of the issue but it seems design similar to yours are working correctly (ex the Digilent Arty has a very similar config: 100MHz sys_clk_freq, RMII PHY, 100Mbit full duplex). Could you provide a way to observe the issue so that I investigate on the Digilent Arty? Thanks.

rprinz08:
Thx for the answer. I will create an example design sohwing the problem (in the next weeks). It actually happens when capturing longer (>1000 bytes) packets into a memory for further processing.

enjoy-digital:
Thanks @rprinz08 for this.

This PR has since been superseded on current master. The MAC/core CDC depth parameters are now configurable through `tx_cdc_depth` / `rx_cdc_depth`, with optional `tx_cdc_buffered` / `rx_cdc_buffered` as well.

This is exposed through `LiteEthMACCore`, `LiteEthMAC`, `LiteEthIPCore` / `LiteEthUDPIPCore`, and `liteeth_gen` YAML configs.

The related integration happened later in: https://github.com/enjoy-digital/liteeth/commit/641c5dbdc77b47e5f31e5627946d27b7607d4150

So I think we can close this older PR as superseded.

PR #82, opened 2021-09-29
mac/sram: Handle endianness only in sram

david-sawatzke:
The hardware udp/ip stack with the expanded (de-)packetizer will only work with LE data and the ports for udp already expect LE data.

Makes more sense to only handle endian in the SRAM.

Thank you @enjoy-digital for taking a look at the previous PRs!


enjoy-digital:
Thanks for the simplificaitons @david-sawatzke, this indeed makes sense. This is merged.

ozel:
it should be noted somewhere in the LiteEth Readme or the LiteX Wiki that the Etherbone / UDP core stores the packet payload in little endian format. It also forwards it like that to streams coming out of `self.ethcore_etherbone.udp.crossbar.get_port()`, getting eventually consumed in little endian format by modules like LiteEthUDPStreamer

PR #83, opened 2021-10-11
frontend/etherbone: Add last_be everywhere

david-sawatzke:
This has been tested with the (De-)Packetizer PR https://github.com/enjoy-digital/litex/pull/1008 and needed in combination with a 32 bit udp/ip stack.

It would be great to get the (De-)Packetizer merged as well, since it seems to work well and is necessary for a 32 bit udp/ip stack.


enjoy-digital:
Thanks @david-sawatzke, this is merged.

PR #84, opened 2021-10-13
core/arp: assert last_be only on the last data word

lschuermann:
This is in line with other components of the LiteEth repository using the `last_be` data qualifier. It is unexpected for last_be to be
asserted on any data word other than the last (i.e. only when `last` is also asserted). In particular it can confuse the Packetizer and cause it to pass through `last_be` on words other than the last as well. This then irritates some parts of the MAC pipeline.

With these changes, ARP works on a 64-bit wide data path (with the {Dep,P}acketizer changes integrated into LiteX).

### DIFF
--- a/liteeth/core/arp.py
+++ b/liteeth/core/arp.py
@@ -51,7 +51,11 @@ class LiteEthARPTX(Module):
         )
         self.comb += [
             packetizer.sink.last.eq(counter == (packet_words - 1)),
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @lschuermann, this is merged. I think we should do the opposite: Make sure the module using `last_be` does not take it into account when last is not set (This generally simplifies things), but since we'll rework the Data Qualifier, let's merge and go further to have a working implementation we'll use for the rework.

PR #85, opened 2021-10-13
liteeth/mac: support SRAM/Wishbone operation with dw > 32 bit

lschuermann:
These is required to support 64-bit wide data paths using the automatic Wishbone bus width conversion as discussed in #75. Otherwise, if the MAC data width deviates from the 32-bit default set previously, received and transmitted packets will be broken.

This also removes an assertion of the MAC data width for the Wishbone interface mode of operation.


enjoy-digital:
Thanks @lschuermann, this is merged.

PR #87, opened 2021-10-27
mac/init: Add 32 & 64 bit support to hybrid mac

david-sawatzke:
Only tested in simulation, might make more sense to add a separate width/cd converter that would be used for the hybrid mac, udp ports and the mac


david-sawatzke:
44b9920 adds a separate converter module, maybe also convenient for the mac pipeline and the udp port, that automagically converts the width in the right clock domain fills LastBe if needed.

enjoy-digital:
Thanks @david-sawatzke, happy to merge if you are confident it does not break 8-bit support. Could you list the tests you are running for this PR here: https://github.com/enjoy-digital/liteeth/issues/89? Thanks.

david-sawatzke:
I've only tested this in simulation, so I'm not *that* confident (same reason it's still a draft). Since you added dw==8 support in sram.py (https://github.com/enjoy-digital/liteeth/commit/1b25f7781c288b3f4376272abc44772e2d033fef) it's probably possible to skip the conversion in `mac/__init__.py` for all `dw`s, but I haven't tried that yet.

david-sawatzke:
Deprecated for in favor of #90

PR #88, opened 2021-10-28
{De,P}acketizer: fix source.last assignment between the two FSMs

lschuermann:
This appears to fix some errors encountered during sending and receiving of packets where sink.valid dropped to zero immediately after last and the last_be overflowed onto the current bus word.

It turns out that the last_be FSM's changes to source.last are not respected by the main FSM due to the way assignments to registers in procedural blocks work on Verilog.

This introduces a selector for which last signal to use, goverened by the last_be FSM.

This has been debugged in collaboration with @david-sawatzke.

### DIFF
--- a/liteeth/packet.py
+++ b/liteeth/packet.py
@@ -43,6 +43,10 @@ class Packetizer(Module):
         if header_words != 1:
             self.sync += If(sr_shift, sr.eq(sr[data_width:]))
 
+        source_last_a = Signal()
+        source_last_b = Signal()
+        source_last_s = Signal()
+
         # FSM.
         self.submodules.fsm = fsm = FSM(reset_state="IDLE")
         fsm_from_idle = Signal()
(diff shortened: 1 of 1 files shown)


david-sawatzke:
We're both sure it worked not too long ago, but I couldn't get it working again. It's *very* weird and I don't quite understand how/why this works/doesn't.

Broken and now working on verilator versions 4.214 and 4.210.

enjoy-digital:
Thanks @lschuermann, @david-sawatzke: Driving directly the same signal from different FSMs should indeed be avoided. Adding a control signal to indicate which FSM is force drive from LastBE's FSM seems indeed a good option here. (As discussed, we'll probably rework this later by avoiding LastBE here). Happy to merge, could you just document the test in https://github.com/enjoy-digital/liteeth/issues/89? Thanks.

david-sawatzke:
A simple crossbar with dw=32 test works on the colorlight board for ping & arp. Seems to work!!

enjoy-digital:
Great, thanks @lschuermann, @david-sawatzke and @jersey99!

PR #90, opened 2021-11-01
mac/init: Add 32 & 64 bit support to hybrid mac (take two)

david-sawatzke:
Deprecates #87 

Tested only in simulation for dw=8 and dw=32 for the whole mac & hardware stack, but *shouldn't* introduce any new failure points.


enjoy-digital:
@david-sawatzke: Thanks! From a first quick look, I really like the fact that it's a lot simpler than the previous solution and also simplifying the existing code. I'll do a closer review very soon.

david-sawatzke:
I've tested this on the colorlight 5a-75b (together with #88) with dw==32 and it works, but I can't really test dw==8 on hardware.

enjoy-digital:
@david-sawatzke: Great, the PR is merged. Would you mind sharing the modification/command used on test it on the color-light in https://github.com/litex-hub/litex-boards/issues/40? Thanks.

PR #91, opened 2021-11-04
WIP: Multiple eth clk domains

jersey99:
Along the lines for having multiple ethernet cores, would mean multiple eth_rx/tx clockdomains.

@enjoy-digital would something like this make sense to you? This way, each phy contains the names of the clock domains and mac core has access to them through the phy?

Note I only did this for the XGMII case at the moment, but I can fix the other files with seemingly no regressions if you think it makes sense.

### DIFF
--- a/liteeth/mac/core.py
+++ b/liteeth/mac/core.py
@@ -37,8 +37,12 @@ class LiteEthMACCore(Module, AutoCSR):
             cd_rx       = "sys"
             datapath_dw = core_dw
         else:
-            cd_tx       = phy.cd_eth_tx
-            cd_rx       = phy.cd_eth_rx
+            if type(phy.cd_eth_tx) is str:
+                cd_tx       = phy.cd_eth_tx
+                cd_rx       = phy.cd_eth_rx
+            else:
+                cd_tx       = phy.cd_eth_tx.name
+                cd_rx       = phy.cd_eth_rx.name
             datapath_dw = phy_dw
         if isinstance(phy, LiteEthPHYModel):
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @jersey99, I generally have a preference to use a `ClockDomainsRenamer` on top of the modules for this. Are you sure the changes you need are not already covered by https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc.py#L1431-L1434?

enjoy-digital:
Sorry @jersey99 for the delay. I just made sure the `add_ethernet`/`add_etherbone ` methods will work with different Ethernet PHYs.

The clock domains are automatically renamed during the integration with:

```python3
        ethmac = ClockDomainsRenamer({
            "eth_tx": phy_cd + "_tx",
            "eth_rx": phy_cd + "_rx"})(ethmac)
```
so your changes are not required with this.

Here is an example with 2 Ethernet PHYs/MACs in the same design:
```python3
self.submodules.ethphy0 = LiteEthPHYMII(
    clock_pads = self.platform.request("eth_clocks", 0),
    pads       = self.platform.request("eth", 0))
self.submodules.ethphy1 = LiteEthPHYMII(
    clock_pads = self.platform.request("eth_clocks", 1),
    pads       = self.platform.request("eth", 1))
self.add_ethernet(name="ethmac0", phy=self.ethphy0, phy_cd="ethphy0_eth")
self.add_ethernet(name="ethmac1", phy=self.ethphy1, phy_cd="ethphy1_eth")
```

PR #92, opened 2021-11-05
Small mac improvements

david-sawatzke:
Two smaller details that I've noticed while debugging etherbone (afaict when receiving a probe, the tx path gets blocked because arp needs to proceed, but the rx path is still blocked by the etherbone request, resulting in a deadlock. fix coming soon™, still need to test on hardware).


enjoy-digital:
Thanks @david-sawatzke!

PR #93, opened 2021-11-06
frontend/etherbone: Move PacketFifo forward in rx chain to also catch probe packets

david-sawatzke:
Otherwise, depending on the caching in the path, a deadlock might occur due to the rx path being blocked by the tx path, being blocked by the arp request, for which the answers get dropped in the rx path.

With this, etherbone works on a hybrid mac on a colorlight 5a-75b using this patch: https://gist.github.com/david-sawatzke/835fa4b8bbd1679f6befd91ac2be7047 and this command `./colorlight_5a_75x.py --board 5a-75b  --with-etherbone --revision 6.1  --sys-clk-freq 40e6 --csr-csv=csr.csv --build --load --uart-name=crossover`. Terminal, ident and reg dumping work

The critical path is quite long (through the whole mac->ip->udp->etherbone->udp->ip->mac chain) so the sys clk target doesn't get higher than 40MHz, but it's probably easily solvable using a few choice buffers.

### DIFF
--- a/liteeth/frontend/etherbone.py
+++ b/liteeth/frontend/etherbone.py
@@ -182,52 +182,46 @@ class LiteEthEtherboneRecordDepacketizer(Depacketizer):
 
 
 class LiteEthEtherboneRecordReceiver(Module):
-    def __init__(self, buffer_depth=4):
+    def __init__(self):
         self.sink   = sink   = stream.Endpoint(eth_etherbone_record_description(32))
         self.source = source = stream.Endpoint(eth_etherbone_mmap_description(32))
 
         # # #
 
-        self.submodules.fifo = fifo = PacketFIFO(eth_etherbone_record_description(32),
-            payload_depth = buffer_depth,
-            param_depth   = 1,
-            buffered      = True
-        )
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks a lot @david-sawatzke for the analysis and PR! The proposed solution is valid but:
- Now does the buffering on a larger group of signals: eth_etherbone_packet_user_description(32) vs eth_etherbone_record_description(32),
- Breaks the symmetry between TX/RX Etherbone Record paths.

I have a preference to just add a 1-Slot PacketFIFO directly in the Probe module as proposed in https://github.com/enjoy-digital/liteeth/pull/94. Is is fine for you? Could you do a test on your hardware with it?

david-sawatzke:
My thoughts were that that since both `Record` & `Probe` needed a Buffer that it might be cheaper & more obvious to combine it, but I haven't verified that it's better in any real way. #94 is also fine for me, both seem good to me.

enjoy-digital:
Thanks. For the simplicity and better decoupling, I have a preference for #94 now so will merge it. We could revisit this in the future if resource usage is not optimal.

PR #94, opened 2021-11-08
frontend/etherbone: Add 1-Slot PacketFIFO on LiteEthEtherboneProbe to prevent deadlock situation.

enjoy-digital:
Based on  @david-sawatzke's analysis:
"Otherwise, depending on the caching in the path, a deadlock might occur due to the rx path being
blocked by the tx path, being blocked by the arp request, for which the answers get dropped in the
rx path."

### DIFF
--- a/liteeth/frontend/etherbone.py
+++ b/liteeth/frontend/etherbone.py
@@ -139,14 +139,21 @@ class LiteEthEtherboneProbe(Module):
 
         # # #
 
+        self.submodules.fifo = fifo = PacketFIFO(eth_etherbone_packet_user_description(32),
+            payload_depth = 1,
+            param_depth   = 1,
+            buffered      = False
+        )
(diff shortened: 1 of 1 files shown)


david-sawatzke:
I've tested this and #93 on the colorlight 5a-75b, but compared to a few days ago now after a few lines of the bios crossterm print on the `ROM` line, a response packet gets lost and litex_server hangs completely. After restarting litex_server & litex_term, the console output continues as expected. Maybe my USB 3.0 gigabit ethernet adapter is interfering, but I doubt it.

Unfortunately this doesn't occur in simulation, so I can't (easily) find out what goes wrong. 

The behaviour of both PRs seems to be completely identical and both `regs` & `ident` work without issue.

enjoy-digital:
Thanks @david-sawatzke, I'm merging it and will also do more testing.

PR #95, opened 2021-11-17
phy/xgmii: handle IFG insertion in PHY, support deficit idle count

lschuermann:
This has a corresponding LiteX PR: enjoy-digital/litex#1100

## Issue

Because XGMII only allows start of frame characters to be placed on lane 0 (first octet in a 32-bit XGMII bus word), when a packet's length % 4 != 0, we can't transmit exactly 12 XGMII idle characters inter-frame gap (the XGMII end of frame character counts towards the inter-frame gap, while start of frame does not). Given we are required to transmit a minimum of 12 bytes IFG, it's allowed to send packet length % 4 bytes additional IFG bytes. However this would waste precious bandwidth transmitting these characters.

Thus, 10Gbit/s Ethernet and above allow using the deficit idle count mechanism. It allows to delete some idle characters, as long as an average count of >= 12 bytes IFG is maintained. This is to be implemented as a two bit counter as specified in IEEE802.3-2018, section four, 46.3.1.4 Start control character alignment.

In practice, the previous implementation of the LiteEthPHYXGMIITX made these issues even more prevalent: because the internal stream interface is 64-bit wide and stream transactions always start aligned to the first octet in a bus word, the previous primitive TX implementation always started transmission on the first octet in the 64-bit XGMII bus word. The IFG inserter operated independently if the PHY and thus made sure to maintain 12 bytes of IFG on the 64-bit stream bus. This means that in a worst case scenario, the IFG could grow to 23 octets. In applications such as Ethernet switches, the consequences would be frequent buffer overruns or corrupt transmissions.

## Changes outlined

Hence this commit introduces a IFG inserter in the LiteEthPHYXGMIITX module itself. It is significantly more complex compared to the gap inserter, but inserts the smallest legal gap as defined by IEEE802.3. Furthermore, it optionally implements the deficit idle count algorithm as described by Eric Lynskey of the UNH InterOperability Lab1 to achieve an average IFG of 12 bytes.

## TODO

This is missing some tests. I have written a test framework which can insert XGMII bus states based on CSV files produced by the Xilinx Vivado ILA and some sample captures between proper 10Gbit/s NICs. I'm planning to have these tests cleaned up and ready in a few days.


lschuermann:
@enjoy-digital I've added the CSV XGMII bus capture based tests (have been the most thorough out of all the ones I've built and uncovered the most edge cases). I think this should be ready now. I can add some additional stream-based tests with pseudorandom data in the next days, either as part of this or a follow-up PR.

enjoy-digital:
Wonderful work @lschuermann, thanks for making it very clear and document it! This is merged.

PR #96, opened 2021-11-21
phy/xgmii: remove vcd dump in test, fix comment

lschuermann:
Removes a slightly inappropriately named VCD dump created by the XGMII tests which may or may not have been a result of some frustration during the test implementation. :)

Also fixes a comment about IFG insertion in the PHY.

@enjoy-digital I swear I've looked over the code multiple times and didn't notice this, it wasn't intentional to be left in there. I did have a good laugh at myself to see this made it into upstream though. :laughing:


enjoy-digital:
@lschuermann, no problem; I can understand... :)

PR #97, opened 2021-12-03
core/udp: Add dw>8 support to LiteEthUDPRX

david-sawatzke:
Pass through last_be or create a new one based on the length if needed

Initialize count at dw//8 and don't subtract source_length since it
could lead to underflow for very short packets. In case source.length is
not dw aligned, count will not be equal, so adjust comparison for that.

Crib last_be Case lookup from mac/sram.py

Only tested with some custom receiver code & dw==32 & in simulation, with correct udp length.

### DIFF
--- a/liteeth/core/udp.py
+++ b/liteeth/core/udp.py
@@ -180,7 +180,7 @@ class LiteEthUDPRX(Module):
         count = Signal(16)
         self.submodules.fsm = fsm = FSM(reset_state="IDLE")
         fsm.act("IDLE",
-            NextValue(count, 0),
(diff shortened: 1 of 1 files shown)


lschuermann (approved):
This (new :smile:) version, along with some other fixes I'll push in a bit, does make Etherbone usable on a 64-bit data path. I didn't test it on other `dw`s though. Thanks @david-sawatzke!

enjoy-digital:
Thanks @david-sawatzke, @lschuermann. Sorry for the delay, this is merged.

PR #98, opened 2021-12-27
icmp: only send a ping reply to type 8 (ping request) packets

michael-betz:
* otherwise liteeth would send a ping reply to `destination unreachable` packets too, which is not wanted.
  * tested in simulation and on hardware (Kintex7)


enjoy-digital:
Thanks @yetifrisstlama! This is merged.

PR #99, opened 2022-01-10
stream: allow LiteEthUDPStreamer to be used as RX only

fjullien:
May be LiteEthUDP2StreamRX shouldn't check the ip_address at all ?

I did not try to use the TX part after this change.

### DIFF
--- a/liteeth/frontend/stream.py
+++ b/liteeth/frontend/stream.py
@@ -68,13 +68,16 @@ class LiteEthUDP2StreamRX(Module):
 
         # # #
 
-        ip_address = convert_ip(ip_address)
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @fjullien, it seems the PR is integrating different changes, can you limit it to making` ip_address` filtering optional on LiteEthUDP2StreamRX?

enjoy-digital:
@fjullien: I just had to use it on a design and did the modification directly in https://github.com/enjoy-digital/liteeth/commit/8faba46517d657b2d732b7d4bf3e10f75c09752b.

PR #102, opened 2022-02-20
phy/a7_1000basex: expose rx/tx polarity controls

sergachev:
This is required on boards with swapped polarities.


enjoy-digital:
Thanks @sergachev, this is merged.

PR #104, opened 2022-04-05
frontend/stream/LiteEthUDPStreamer: make send_level configurable

hcab14:
When `send_level == 1`, UDP payload size is `data_width` which limits throughput.

Please check carefully because I am new to using litex/liteeth.


enjoy-digital:
Hi @hcab14,

Thanks for the PR and sorry for the delay. `send_level` has been removed with https://github.com/enjoy-digital/liteeth/commit/6cf7759c9bfaf221eec6e62cae504c7e88ab690b you can now just use packet delimiters to send packets of specific length.

PR #106, opened 2022-04-26
Etherbone testing

michael-betz:
Various tweaks to improve icmp and etherbone testing and troubleshooting.

  * icmp header: split `quench` field into `ident` and `sequence` fields  
  * arp.py: make the cached mac address writable from a testbench (to speed it up)
  * model/etherbone.py: add timeout to receive(), fix length field in header
  * model/icmp.py: make `send()` calculate the checksum of the icmp-packet
  * model/ip.py: fix checksum calculation for odd number of bytes
  * model/phy.py: add optional check for stalling (valid de-asserted mid-stream) because some PHYs don't tolerate that. 
  * dump sent / received packets to a .pcap file for analysis with wireshark
  * model/udp.py: fix total_length field in header
  * test_etherbone.py: check probe, write and read response for 8, 32 and 64 bit datapaths
  * test_icmp.py: check ping response for 8, 32 and 64 bit datapaths

This needs some changes to litex, in particular to `PacketLogger`, `PacketStreamer` and `StrideConverter`, which is why CI is failing. See the [corresponding pull request to litex](https://github.com/enjoy-digital/litex/pull/1283)


enjoy-digital:
Thanks again @michael-betz for the original work here.

This PR had become a bit old and no longer applied cleanly to current master. The companion LiteX-side work was also handled separately: the still-relevant StrideConverter / last_be down-conversion fix landed in LiteX here:

https://github.com/enjoy-digital/litex/commit/a2a62dcc662465015dcd55d0de701d41de40a660

I created a fresh LiteEth PR from current master here:

https://github.com/enjoy-digital/liteeth/pull/205

It keeps the still-useful ideas from this PR, with credit in the commit:
- fixes ICMP/UDP/IP test-model checksum and length handling;
- makes the PHY test model stream bytes correctly on wider datapaths;
- adds active ICMP echo and PHY model regressions;
- replaces the old hanging Etherbone full-stack test with a faster UDP-boundary Etherbone frontend regression covering probe/write/read behavior.

So I think we can treat this older PR as superseded by #205.

PR #115, opened 2022-06-29
LiteEthMAC: remove logic loop in hybrid mode

Xiretza:
The hardware and cpu ports' valid signals can not be made dependent on their own ready signals, since this creates combinatorial loops. This should be unnecessary anyway though, the only thing that matters is that the valid signal is not asserted while the other sink is not ready.

### DIFF
--- a/liteeth/mac/__init__.py
+++ b/liteeth/mac/__init__.py
@@ -84,7 +84,6 @@ class LiteEthMAC(Module, AutoCSR):
 class LiteEthMACCoreCrossbar(Module):
     def __init__(self, core, crossbar, interface, dw, hw_mac=None):
         rx_ready = Signal()
-        rx_valid = Signal()
 
         # IP core packet processing
         self.submodules.packetizer   = LiteEthMACPacketizer(dw)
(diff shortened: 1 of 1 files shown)


Xiretza:
I haven't had time to actualy test this yet, so I'll keep it a draft for now unless you're sure this is correct.

enjoy-digital:
Thanks @Xiretza, I'll study it in the next days.

enjoy-digital:
Thanks again @Xiretza for this.

This was still relevant in principle, but the hybrid MAC code has changed enough since the original PR that the old patch no longer applied cleanly, especially with the current `hw_mac` filtering path.

I created a fresh PR from current master here:

https://github.com/enjoy-digital/liteeth/pull/206

It preserves the current routing semantics:
- local unicast -> hardware path only;
- broadcast/multicast -> hardware + CPU paths;
- other unicast -> CPU path only.

The refreshed version avoids driving a sink `valid` from that same sink `ready`, and adds focused regressions for both the no-filter broadcast fanout and the filtered local-unicast path.

So I think we can treat this older PR as superseded by #206.

PR #116, opened 2022-06-30
core: expose hybrid MAC mode

Xiretza:
This allows using both the hardware IP/UDP stack as well as a CPU attached directly to the MAC at the same time.

### DIFF
--- a/liteeth/core/__init__.py
+++ b/liteeth/core/__init__.py
@@ -15,6 +15,7 @@ from liteeth.core.icmp import LiteEthICMP
 
 class LiteEthIPCore(Module, AutoCSR):
     def __init__(self, phy, mac_address, ip_address, clk_freq, dw=8,
+        hybrid            = False,
         with_icmp         = True,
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @Xiretza, this looks good. I'll study it in the next days to see if we use this name for the parameter.

enjoy-digital:
Thanks again @Xiretza.

This idea was valid, but the original patch is now superseded. Current master exposes a more general `interface` parameter on `LiteEthIPCore` / `LiteEthUDPIPCore`, so users can now request this with:

  ```python
  LiteEthUDPIPCore(..., interface="hybrid")
  ```

This was added in daf1a1ac63ea0e610e8063112fdbfdcfb17c46f1.

Also, the related hybrid MAC valid/ready issue has now been refreshed separately in #206.

One adjacent improvement still possible would be exposing this through the standalone liteeth_gen YAML flow, but that would be a separate PR from this old constructor API change.

So I think we can close this as superseded.

PR #119, opened 2022-09-06
liteeth/gen: Use hw_init_reset for rgmii

mkj:
I'm not sure of the original reason for this being marked FIXME, so if the `hw_init_reset` needs to remain I can instead make this a configuration option for `litedram_gen`. 

This re-enables hw_init_reset that was disabled in
ea24ff6 ("liteeth_gen:  improve readability and add clk_freq checks.")

A chip reset is necessary using the Microchip KSZ9031RNX
PHY on an antmicro-artix-dc-scm board (possibly depending on startup
power timing)

### DIFF
--- a/liteeth/gen.py
+++ b/liteeth/gen.py
@@ -204,7 +204,7 @@ class PHYCore(SoCMini):
                 pads               = platform.request("rgmii_eth"),
                 tx_delay           = core_config.get("phy_tx_delay", 2e-9),
                 rx_delay           = core_config.get("phy_rx_delay", 2e-9),
-                with_hw_init_reset = False) # FIXME: required since sys_clk = eth_rx_clk.
+            )
         else:
             raise ValueError("Unsupported PHY")
         self.submodules.ethphy = ethphy


enjoy-digital:
Hi @mkj,

sorry for the delay. The FIXME was probably here when used with the HW stack, I'll have a closer look at the use-case that was requiring it.

enjoy-digital:
Thanks again @mkj.

This use case is still relevant: current master still forces `with_hw_init_reset=False` for generated S7/ECP5 RGMII cores, so `liteeth_gen` users cannot request the startup PHY reset pulse.

I created a fresh PR from current master here:

https://github.com/enjoy-digital/liteeth/pull/207

The refreshed version exposes this as an explicit YAML option:

```yaml
phy_hw_init_reset : True
```

It keeps the current default as `False`, since the old inline FIXME suggested some generated-core setups needed the reset pulse disabled. This should cover the KSZ9031RNX / board power-up timing use case without changing behavior for existing generated RGMII cores.

So I think we can treat this older PR as superseded by #207.

PR #121, opened 2022-11-03
LiteEthStream2UDPTX: fix for end of buffer tags

hcab14:
Please have a look. For it works, _i.e._ I can use the entire GBit ethernet bandwidth (using a butterstick).

I wonder how one could make a CI test for this.

### DIFF
--- a/liteeth/frontend/stream.py
+++ b/liteeth/frontend/stream.py
@@ -30,35 +30,60 @@ class LiteEthStream2UDPTX(Module):
             level   = Signal(max=fifo_depth+1)
             counter = Signal(max=fifo_depth+1)
 
-            self.submodules.fifo = fifo = stream.SyncFIFO([("data", data_width)], fifo_depth, buffered=True)
+            # intermediate buffer used to scan for fifo.source.last tags
+            self.submodules.fifo_buffer = fifo_buffer = stream.SyncFIFO([("data", data_width)], fifo_depth, buffered=True)
+
+            # sink FIFO
+            self.submodules.fifo        = fifo        = stream.SyncFIFO([("data", data_width)], fifo_depth, buffered=True)
             self.comb += sink.connect(fifo.sink)
 
             self.submodules.fsm = fsm = FSM(reset_state="IDLE")
             fsm.act("IDLE",
-                # Send FIFO contents when we have a full-packet or when FIFO is full.
-                If((fifo.source.valid & fifo.source.last) | ~fifo.sink.ready,
+                #  wail until the FIFO is full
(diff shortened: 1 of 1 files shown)


rowanG077:
Why wasn't it possible to reach GBit ethernet speeds before this? If I read the code correctly this will double the bram usage. Which I'm not a fan of.

hcab14:
It was possible to reach GBit speeds before but the `last` tags were not properly dealt with.

In two cases it did work as expected:
1. `last=0`: the FIFO was always fully filled and then its contents were sent in one UDP packet
2. `last=1`: FIFO content was sent one by one

But, _e.g._, if I wanted to send UDP packets with two FIFO entries, _i.e._, `last` was made to toggle it did not work because `last` is checked only in the `IDLE` state and not during the `SEND` state.

rowanG077:
I understand. I had this issue too.

But I think double buffering is not the correct path here due to such waste of resources. I dealt with it by building a custom FIFO and using the UDP raw interface.What I would advise here is to replace the `SyncFIFO` with the [`PacketFIFO`](https://github.com/enjoy-digital/litex/blob/master/litex/soc/interconnect/packet.py#L364). It should be possible to send full packets only. I didn't go this route because I'm not proficient in migen/litex.

The drawback is that the user will have to specify the length of the packet up front.

enjoy-digital:
Thanks @hcab14, @rowanG077, while discussing with @hcab14 about the issue (on IRC IIRC), I also thought about replacing the `SyncFIFO `with a proper `PacketFIFO`. I'll try to have a closer look at this soon. We should indeed try to minimize resource usage.

enjoy-digital:
Thanks again @hcab14.

This was still a real issue on current master: `LiteEthStream2UDPTX` FIFO mode still only looked at the input-side `last` while the TX FSM was in `IDLE`, so back-to-back multi-word packets could leave later packet boundaries queued without being emitted.

I created a fresh focused PR from current master here:

https://github.com/enjoy-digital/liteeth/pull/209

The refreshed version keeps the scope focused and uses packet-aware buffering:
- replaces the TX FIFO/FSM path with `PacketFIFO`;
- preserves queued packet boundaries behind an active transmission;
- stores payload data plus per-packet length, while latching IP/UDP port when a queued packet becomes active on TX;
- keeps the legacy behavior of emitting a packet when a no-`last` stream reaches `fifo_depth`;
- keeps the current disable-mid-packet behavior of finishing the visible packet before stopping.

It also adds focused regressions for the back-to-back multi-word packet case and the FIFO-depth/no-`last` case.

So I think we can treat this older PR as superseded by the new one, with credit preserved there.

PR #124, opened 2022-12-09
wishbone rx data corruption

sensille:
See commit message, this fixes a data corruption in the RX path. It also fixes a typo that prevented having a different number of rx and tx buffers.


enjoy-digital:
Thanks @sensille! This looks good and has been merged. Sorry for the delay.

PR #126, opened 2023-01-23
option to set LAN8720 mode during reset

machdyne:
I'm working on a [custom LAN8720A ethernet PHY PMOD](https://github.com/machdyne/katze) designed specifically for LiteX. I discovered that the configuration straps need to be set in a specific way during reset to select the correct mode.

This PR allows you to set the LAN8720A mode during reset, for example:

```
self.submodules.ethphy = LiteEthPHYRMII(
	clock_pads = platform.request("eth_clocks"),
	pads = platform.request("eth"),
	with_hw_init_reset=True,
	hw_init_mode_cfg=[1,1,1],		# "All capable. Auto-negotiation enabled."
	refclk_cd=None)
self.add_ethernet(phy=self.ethphy)
```

Please let me know if there's a better way to do this.

Thanks.

### DIFF
--- a/liteeth/phy/rmii.py
+++ b/liteeth/phy/rmii.py
@@ -43,7 +43,7 @@ class LiteEthPHYRMIITX(Module):
 
 
 class LiteEthPHYRMIIRX(Module):
-    def __init__(self, pads):
+    def __init__(self, pads, hw_init_mode_cfg):
         self.source = source = stream.Endpoint(eth_phy_description(8))
 
         # # #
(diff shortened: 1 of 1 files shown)


ozel:
The Waveshare 8720A PMOD ethernet board has precisely the same problem. It needs to be plugged angled - so power connects before the RMII pins - otherwise, it will end up in unuseful 10 Mbit half duplex mode.

While I am also curious about the best way to pre-configure pins before specific submodules are loaded, I came to the conclusion there are better fixes for this particular issue. There should be additional 10k pull-up resistors on the pmod itself (RX0, RX1, CRS), like in this design:
https://github.com/swetland/ethernet-pmod/blob/main/export/ethernet-pmod-v2-schematic.pdf 

Alternatively, one can try enabling pull-ups on the FPGA side. The ECP5 appears strong enough to drive the 8720A pins on the Waveshare pmod after adding `Misc("PULLMODE=UP")` to the three mode pins.
I've also added `Misc("SLEWRATE=FAST")` to the whole set of RMII pins, following the `sdcard` io configuration in `litex_boards.platforms.radiona_ulx3s`. However, I'm not sure if the latter has an effect on the pull-up timing/strength.

machdyne:
Thanks, the ECP5 pull-ups are a much simpler solution and it seems to be working. Will probably add those resistors in the next board revision.

PR #136, opened 2023-07-06
Add gateway support

rowanG077:
This is a simple method to add a static gateway + netmask.

TODO: Test in hardware if you can actually go over the internet with this.

I also added UDPRawCore in the generator.


rowanG077:
I think the logic now is correct. I will test out whether it works in Hardware soon.

liteeth/core/ip.py
```diff
@@ -145,7 +158,21 @@ def __init__(self, mac_address, ip_address, arp_table, dw=8):
...
         )
-        self.comb += arp_table.request.ip_address.eq(sink.ip_address)
+
+        if gateway is not None:
+            masked_ip = Signal(32, reset_less=True)
+            in_subnet = Signal(1, reset_less=True)
+
+            self.sync += masked_ip.eq(sink.ip_address & netmask)
+            self.sync += in_subnet.eq(masked_ip == netaddress)
+            self.sync += If(in_subnet,
```
rowanG077, line 168: @enjoy-digital  Now having slept on it, I'm not actually sure this will work correctly. Essentially the IP address now needs to be stable for 3 clock cycles before the `SEND_MAC_ADDRESS_REQUEST` is reached. I don't know enough about packetizers and entire stack whether this is true. We have a single cycle for sure so I could shrink the pipeline.

enjoy-digital:
@rowanG077 Thanks for the PR, I'll look at it. It's however integrating 3 different changes. Would you mind creating a PR per specific changes? This will ease review and merging progressively the different features.

For the UDP Raw mode, it would probably be better to add a mode parameter to the defined UDP ports, where mode could be `stream` or `raw` and then implement the right logic in the generator. This would even be more flexible since some ports could use be in  `stream`  while others in  `raw` mode.

PR #138, opened 2023-07-14
phy/ecp5rgmii.py: Add support for dynamic link speeds

rowanG077:
This is still a draft. It doesn't work yet but if you have time to look at it the general structure should be finalized.

There are 3 main changes to make this work:

- 1G links use DDR but 100M and 10M do not. Based on the in-band status we use back-pressure on the TX side to stall incoming bytes while the nibbles get transmitted out.
- On the RX side use the rising edge DDR inputs to capture the incoming nibble. Use an 8-bit shift register to shift in the nibble on every data valid. Alternate outgoing valid on every data valid to ensure the full byte is transmitted into the core.
- There are two configs that can be used. Either use the RX clock as the TX clock (loop clocking) or provide a 125Mhz clock yourself. The latter should be strongly encouraged but it's not always possible. In the case that a 125Mhz clock is provided a divided down clock to 25Mhz or 2.5Mhz is created inside the Phy depending on link speed just for the purpose of going into the clock forwarding ODDR primitives. This creates a clock enable signal for the TX side so it knows when it needs to present the next nibble. In the case of loop clocking no clock division is done since the RX clock is already at the correct speed.

### DIFF
--- a/liteeth/mac/core.py
+++ b/liteeth/mac/core.py
@@ -8,6 +8,8 @@
 # Copyright (c) 2023 LumiGuide Fietsdetectie B.V. <goemansrowan@gmail.com>
 # SPDX-License-Identifier: BSD-2-Clause
 
+import math
+
 from liteeth.common import *
 from liteeth.mac import gap, preamble, crc, padding, last_be
 from liteeth.phy.model import LiteEthPHYModel
@@ -103,7 +105,14 @@ class LiteEthMACCore(Module, AutoCSR):
                 self.pipeline.append(tx_preamble)
 
             def add_gap(self):
-                tx_gap = gap.LiteEthMACGap(phy_dw)
+                # Some phys, ECP5 for example, have a byte time enable to support
+                # dynamic link speeds.
+                # In the gap inserter we need to ensure this is enable holds for
+                # the gap counter. If not we would insert a gap that might be too
+                # short.
+                default_gap = Constant(math.ceil(eth_interpacket_gap/(phy_dw//8)))
+                gap_len = getattr(phy, "gap", default_gap)
+                tx_gap = gap.LiteEthMACGap(phy_dw, gap_len)
                 tx_gap = ClockDomainsRenamer("eth_tx")(tx_gap)
                 self.submodules += tx_gap
                 self.pipeline.append(tx_gap)
--- a/liteeth/mac/gap.py
+++ b/liteeth/mac/gap.py
@@ -4,35 +4,33 @@
 # Copyright (c) 2015-2021 Florent Kermarrec <florent@enjoy-digital.fr>
 # Copyright (c) 2015-2017 Sebastien Bourdeauducq <sb@m-labs.hk>
 # Copyright (c) 2018 whitequark <whitequark@whitequark.org>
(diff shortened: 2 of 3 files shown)


rowanG077:
I will test this in depth this week. But the basics should work now.

enjoy-digital:
Thanks @rowanG077, that's also a feature I was interested (for ECP5 and the other RGMII PHYs). I'll try to do a review soon.

rowanG077:
I had some troubles with switching. In turns out that the clock switch MUST be glitchless so I added that. 

Dynamic 1G/100M mode works, 10M doesn't.... I really have no clue why. 10M sends the same bytes as 100M. Simulation shows nothing weird. 

I tested both the loop clock scenario and the FPGA generated tx clock. In both scenarios 1G/100M work and 10M does not. Maybe I'm missing something?

enjoy-digital:
Great! 1G/100M will probably be the more useful. I'll look at the code to see if I see an obvious reason for the 10M to not work.

rowanG077:
I think it's due to the IPG inserter running at clock cycle speed whereas now with a dynamic link speed this means the IPG is too small for 100M/10M. I probably just was unlucky 100M works fine. I fixed this. I will try in hardware tomorrow.

rowanG077:
This was not the problem... 10M still completely silent. I'm sure that at the very least the RX path is broken. I don't even receive the requests on the FPGA.

So there are quite a few scenarios, but I could not test them all myself due not having a FPGA board with RGMII in-band status support. loopclocking yes or no, in-band status yes or no and  external linkstate yes or no. Below is the matrix which indicates what config I tested.

```
  loopclock | in-band status | linkstate | Tested (1G/100M working)
      y     |      n         |    y      |    y
      n     |      n         |    y      |    y
      y     |      n         |    n      |    y (1G support only)
      n     |      n         |    n      |    y (1G support only)
      y     |      y         |    n      |    n
      n     |      y         |    n      |    n
```

majianjia:
I met the same problem. Not working in 100M/SDR mode.

rowanG077:
100M should work fine. 10M is the problem. What phy are you using?

majianjia:
> 100M should work fine. 10M is the problem. What phy are you using?

I am trying out colorlight i5 board (Broadcom B50612D). i have was once made it negotiated at 1Gbps but falled down to 100M and never works again. 
This is how i came to here.

I wonder why this pr wasnt merged? Maybe I should just try your work.
Thanks

rowanG077:
Without this PR there is no 10M/100M support only gigabit support. This PR aimed to build 10M/100M support.

I don't know why it wasn't merged. At this point this PR doesn't hold much value for me anymore.

enjoy-digital:
Thanks again for the original work here. The idea is still useful, but this PR had become old/draft and was not directly mergeable on current LiteEth.

I created a fresh PR from current master: #210. It keeps the implementation opt-in and focused:

- derive ECP5 RGMII link speed from in-band status;
- adapt TX/RX for 10/100M SDR nibble operation;
- adapt TX clock forwarding and MAC IFG handling;
- add focused TX/RX/clock tests and generator smoke coverage.

The new commit credits your original PR/ideas with a `Co-authored-by` trailer. Hardware validation at 10/100M is still needed, especially since this original draft mentioned remaining 10M issues, so the new support stays opt-in for now.

PR #137, opened 2023-07-10
gen.py: Add UDP raw mode

rowanG077:

PR #139, opened 2023-07-16
core/ip.py: Add gateway support

rowanG077:
This is not yet exposed in `gen.py`. I will make a PR for that when https://github.com/enjoy-digital/liteeth/pull/137 is merged.


rowanG077:
@enjoy-digital Could you trigger CI again? I don't think the fail is due to it not working. Rather it was cancelled(?) for some reason..

enjoy-digital:
@rowanG077: Sure. It seems to have reached a timeout: 
image

enjoy-digital:
Thanks again @rowanG077.

This is still useful: current master still ARPs directly for the destination IP for all unicast IPv4 packets, so off-subnet routing needs gateway support.

I created a fresh focused PR from current master here:

https://github.com/enjoy-digital/liteeth/pull/208

The refreshed version keeps the default behavior unchanged and adds optional `gateway_ip`/`netmask` support:
- same-subnet destinations ARP for the destination IP;
- off-subnet destinations ARP for `gateway_ip`;
- the IPv4 destination address is left unchanged.

It also exposes the options through `LiteEthIPCore` / `LiteEthUDPIPCore` and `liteeth_gen`, and adds focused ARP target-selection regressions.

So I think we can treat this older PR as superseded by the new one.

PR #141, opened 2023-07-24
gen: add UDP raw listen-port selection

rowanG077:
This rebases Rowan Goemans' original `udpraw` listen-port fix on current master and adds a small cleanup commit on top.

Background: https://github.com/enjoy-digital/litex/pull/1733

## Purpose

`udpraw` should not use the TX-side `sink_dst_port` signal as its RX crossbar selector. The receive/listen port and transmitted destination port are different concepts:

- `udp_port` in the YAML config now selects the static UDP port received by the raw endpoint.
- If `udp_port` is omitted, the generated raw endpoint exposes a dynamic `udp_listen_port` IO.
- `sink_dst_port` remains per-packet TX metadata only.
- `ip_address` is rejected for raw ports since raw mode already carries per-packet IP metadata.

## Credit

- Original fix/idea by Rowan Goemans in the first commit.
- Follow-up cleanup/tests in the second commit.

## Tested

- `python3 -m unittest test.test_gen`
- `python3 -m py_compile liteeth/gen.py test/test_gen.py`
- `git diff --check`

### DIFF
--- a/examples/udp_raw_ecp5rgmii.yml
+++ b/examples/udp_raw_ecp5rgmii.yml
@@ -41,13 +41,14 @@ rx_cdc_buffered : True
 # control of transmitted UDP packet sizes.
 udp_ports:
   raw:
-    data_width : 32
-    mode       : raw
+    data_width: 32
+    udp_port: 2000
+    mode: raw
   streamer1:
-    data_width : 32
-    port       : 1337
-    mode       : streamer
+    data_width: 32
+    udp_port: 1337
+    mode: streamer
   streamer2:
-    data_width : 32
-    port       : 6077
-    mode       : streamer
+    data_width: 32
+    udp_port: 6077
+    mode: streamer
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Rebased this PR on current `master` and added a cleanup commit on top of Rowan's original fix.

The cleaned-up version keeps the original intent but avoids using `sink_dst_port` as the RX selector: raw ports now listen on static `udp_port` when configured, or expose a dedicated dynamic `udp_listen_port` IO when omitted. `sink_dst_port` remains TX packet metadata only.

I also fixed the dynamic-port ordering issue, changed the unsupported raw `ip_address` case to a `ValueError`, and added generator tests for static, dynamic, and invalid raw-port configurations.

Tested with:

- `python3 -m unittest test.test_gen`
- `python3 -m py_compile liteeth/gen.py test/test_gen.py`
- `git diff --check`

PR #148, opened 2023-09-05
core/dhcp: Lease time support

rowanG077:
This PR adds DHCP lease time support. To do that a DHCP OPTIONS parser is implemented.

At first I tried to do it fully in the 32-bit datapath. But it's simply too much. Header can be at any position, Multiple options per 32-bit packet are possible, padding which increases variety even more... If LiteX had some kind of function support It would be possible but I simply don't know how to express that.

So in the end I decided to just bite the bullet and do options parsing with 8-bit width.

Another problem that is fixed now is that `MESSAGETYPE` option no longer is assumed to be first option. The spec does not say it must be the first option so this could be very problematic for some Liteeth DHCP server combinations.

I did already add gateway and netmask parsing but they aren't wired up. 

TODOs:

- [x] Make it function correctly
- [x] Use received lease time to automatically restart DHCP process before the lease has expired
- [x] Make the DHCP module self contained. No longer have external `start`, `timeout` and `done` signals

### DIFF
--- a/liteeth/core/dhcp.py
+++ b/liteeth/core/dhcp.py
@@ -3,19 +3,17 @@
 #
 # Copyright (c) 2023 Florent Kermarrec <florent@enjoy-digital.fr>
 # Copyright (c) 2023 LumiGuide Fietsdetectie B.V.
+# Copyright (c) 2023 Rowan Goemans <goemansrowan@gmail.com>
 # SPDX-License-Identifier: BSD-2-Clause
 
 """
 DHCP
 
-Minimal DHCP (IPV4) support for LiteEth.
+DHCP (IPV4) support for LiteEth.
 
 Limitations/TODOs:
-- No lease time parsing/support, user logic should consider it short (or known from server) and
-issue a DHCP request regularly. Limitations is due to 32-bit data-path and parsing. Switching to a
-8-bit data-path for DHCP options would allow supporting it more easily.
+- Router(gateway) and subnet mask are parsed out of the DHCP options but as of now unused.
 - Additional checks could be made on RX (see FIXMEs, but cost logic on FPGA).
-- Define more DHCP constants and use them in the code.
 """
 
 from migen import *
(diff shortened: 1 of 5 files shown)


liteeth/core/dhcp.py
```diff
@@ -304,11 +313,214 @@ def __init__(self, udp_port):
...
+        self.sink    = sink   = stream.Endpoint(eth_dhcp_opt_description(32))
+
+        data         = Signal(32)
+        byte_en      = Signal(4)
+        latched_be   = Signal(4)
+        latched_last = Signal()
+        last_byte    = Signal()
+
+        self.comb += [
+            # @Florent: Is this necessary? Can we assume last_be is 0b1000 For non-last words?
```
rowanG077, line 331: @enjoy-digital See comment

rowanG077:
@enjoy-digital This works now and is ready for review.

Since the DHCP module can now be self-contained I removed the exposed signals in the core generator.

rowanG077:
@enjoy-digital Just pinging you so it doesn't fly under the radar. I would like it if this can get merged since I really prefer to use the upstream liteeth instead of a mass of custom PR :).

enjoy-digital:
@rowanG077: Thanks for the PR and sorry for the delay. I'll try to plan some time to look at this very soon.

rowanG077:
@enjoy-digital I hate to bother you with this but this and other PRs have been waiting for a while. This means we currently diverge quite a bit from upstream liteeth and I don't really want that. If you want me to put this PR in a different form or walk you through the changes I can do so as well.

enjoy-digital:
Hi @enjoy-digital,

sorry for the delay, lots of different priorities to handle but I'll try to do a closer review soon to avoid blocking things.

enjoy-digital:
Hi @rowanG077,

sorry for the delay, lots of different priorities to handle but I'll try to do a closer review soon to avoid blocking things.

enjoy-digital:
Thanks again for the DHCP lease-time work here. Some parts of this PR are now already covered on master, notably parsing the DHCP message type without assuming it is the first option.

I created a fresh, focused PR from current master for the still-relevant lease-time piece: #211.

It keeps the scope small:

- parse DHCP option 51 (`IP Address Lease Time`) in the existing master DHCP options FSM;
- expose the parsed lease time on `LiteEthDHCPRX`;
- propagate the ACK lease time on `LiteEthDHCP`;
- keep the existing standalone DHCP control API unchanged for now.

The new commit credits your original PR/ideas with a `Co-authored-by` trailer. I left automatic renewal/retry policy and larger generator/DHCP API restructuring out of this first PR so they can be reviewed separately.

PR #149, opened 2023-10-23
core/__init__: LiteEthUDPIPCore, LiteEthIPCore: expose interface & endianness at LiteEthUDPIPCore constructor. LiteEthIPCore: don't hardcode interface, pass macaddr and endianness to LiteEthMAC

trabucayre:
To avoid code dupplication and to be able to have hybrid interface (as siglent_sds1104xe) some parameters must be available at target level


enjoy-digital:
Thanks @trabucayre!

PR #157, opened 2024-03-21
mac/crc.py: optimize CRC validation

rowanG077:
This is a WIP!! Do not merge yet. I need to validate it first.

The idea here is that we can pre-calculate different residues for a zero padded stream. We can just throw the full width into the full width engine, while making sure we set invalid bytes to 0 and then compare with the appropriate check value. This saves us from having to instantiate multiple crc engines on the receive path, which is nice since they are relatively expensive.

I did not do hardware tests yet. I will post resource usage differences later once I have validated it works

### DIFF
--- a/liteeth/mac/crc.py
+++ b/liteeth/mac/crc.py
@@ -92,10 +92,10 @@ class LiteEthMACCRCEngine(Module):
 
 @ResetInserter()
 @CEInserter()
-class LiteEthMACCRC32(Module):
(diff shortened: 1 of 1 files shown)


rowanG077:
By default this is not actually more efficient on yosys. This is due to https://github.com/YosysHQ/yosys/issues/4293 closing because if it does not provide the benefit by default it might as well not be done.

PR #160, opened 2024-04-13
phy/ecp5rgmii.py: In-Band Status CSRField("clock_speed") size fixed

whiteb3ar:
clock_speed size increased to 2 otherwise 1 bit of data is lost


enjoy-digital:
Good catch, thanks @whiteb3ar!

PR #163, opened 2024-07-14
Update a7_gtp.py  transceiver resets have to stay low for 10us.

cyntem:
After testing several boards with XC7A200T chips, I found, that SFP works well only with setting more than 10us. The most of the XC7A200T chips works with 500ns, but some boards need 10us to 100% start work.


enjoy-digital:
Thanks @cyntem, this is merged.

PR #164, opened 2024-07-14
Only import liteiclink when required

m-byte:
As liteiclink is only used in the phy implementations of a few Xilinx/AMD FPGAs, it does not make sense that it would be required to build liteeth for any FPGA.


m-byte:
@enjoy-digital sorry to also bug you with this and PR #165 as well, but it would be great if you could check these out as well.

enjoy-digital:
Thanks @m-byte, this is merged.

PR #165, opened 2024-07-22
MAC: Implement address filtering for logic interface in hybrid mode

m-byte:
With this, in hybrid mode, the MAC address is used to filter packets destined for the logic interface. Packets addressed to either the broadcast MAC, the IPv4 Multicast MACs and the IPv6 Multicast MACs are forwarded to both logic and CPU interface.

### DIFF
--- a/liteeth/mac/__init__.py
+++ b/liteeth/mac/__init__.py
@@ -130,16 +130,24 @@ class LiteEthMACCoreCrossbar(Module):
             ]
 
             # RX packetizer broadcast
+            mac_local = Signal()
+            mac_bcast = Signal()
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @m-byte!

enjoy-digital:
This is merged, we'll have to see if the additional checks don't degrade too much timings, if so, we could make some of them configurable/optional.

enjoy-digital:
@m-byte: I added some comments on top of it to help understand/maintain the filtering part: https://github.com/enjoy-digital/liteeth/commit/bfc07e543aedac0403d024e116623fe010d2daeb

PR #166, opened 2024-08-18
Fix etherbone reads

### LINKED ISSUE #2031: Etherbone issues
I am trying to get etherbone working with my butterstick, but it seems that litex_term and wishbone-tool are both failing to read the uart correctly when using --uart-name=crossover.  All that is read is a constant spew of 0xa resulting in pages of newlines.  I have verified in wireshark using the etherbone dissector that the port read for the uart_xover_rxempty is returning 0 constantly and that the port read for uart_xover_rxtx is constantly reading 0xa.

When I use uartbone instead of etherbone, the reads are correct and the bios terminal is displayed correctly, so I can discount the crossover uart as being the issue.

I'm pretty new to LiteX so I did some digging in LiteEthEtherboneWishboneMaster and compared it to Stream2Wishbone (for uartbone) and they do seem a little divergent in their manipulation of the wishbone bus, but I'm not yet well educated in LiteX's stream logic or wishbone itself to diagnose what the issue might be - it feels like the bus read should latch the data instead of assuming it is always valid after the wishbone ack, but like I said I'm pretty new to this.

I attempted to use litex_sim with etherbone to narrow things down, but the tap0 device created by sim.py is bound to 192.168.1.100 by default and litex_server fails to sucessfully read the initial packet (error is "assert packet.pr == 1").  

Any ideas what might be going on?

Thanks!

m-byte:
Since https://github.com/enjoy-digital/litex/pull/1999, etherbone reads might result in garbage being output. This is caused by `be` not being set during the read.

Fixes https://github.com/enjoy-digital/litex/issues/2031

### DIFF
--- a/liteeth/frontend/etherbone.py
+++ b/liteeth/frontend/etherbone.py
@@ -251,6 +251,7 @@ class LiteEthEtherboneRecordReceiver(LiteXModule):
             source.last.eq(count == fifo.source.rcount-1),
             source.last_be.eq(source.last << 3),
             source.count.eq(fifo.source.rcount),
+            source.be.eq(fifo.source.byte_enable),
             source.base_addr.eq(base_addr),
             source.addr.eq(fifo.source.data[2:]),
             fifo.source.ready.eq(source.ready),


enjoy-digital:
Thanks @m-byte!

PR #169, opened 2024-09-13
phy: rmii: use ClockSignal(refclk_cd) to drive DDROutput

maass-hamburg:
use ClockSignal(refclk_cd) to drive DDROutput.

with this the DDROutput can be used on efinix platforms. There the ClockDomain has to be the one from the PLL. (https://github.com/enjoy-digital/litex/pull/2060)


enjoy-digital:

PR #172, opened 2024-09-25
phy/rmii: fix it for efinix

maass-hamburg:
@enjoy-digital 
On efinix platforms the clk signal of
`SDROutput` and `SDRInput` has to come
from the PLL.

### DIFF
--- a/liteeth/phy/rmii.py
+++ b/liteeth/phy/rmii.py
@@ -18,7 +18,7 @@ from liteeth.phy.common import *
 # LiteEth PHY RMII TX ------------------------------------------------------------------------------
 
 class LiteEthPHYRMIITX(LiteXModule):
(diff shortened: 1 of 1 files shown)


trabucayre:
Tested for compat with nexys4ddr.
Just a small remark: `clk_signal` depends on `refclk_cd` but in both case `cd_eth_rx` & `cd_eth_tx` are connected to the corresponding reference: it's seems more easy/less noisy to use:
- `ClockSignal("eth_tx") in `LiteEthRMIITX`
- `ClockSignal("eth_rx") in `LiteEthRMIIRX`

enjoy-digital:
Thanks @maass-hamburg, it's merged, I'll try to simplify is possible following your suggestions @trabucayre.

PR #173, opened 2024-09-26
phy/trionrgmii,titaniumrgmii: replaces str by ClockSignal for ClkInput and PLL

trabucayre:
With recent commit for efinix toolchain the use of str of ClkInput and PLL clkin is no more required and a clock signal may/must be used instead.


enjoy-digital (approved):
Thanks @trabucayre!

PR #177, opened 2024-12-11
mac/core: Allow using core_dw smaller than phy_dw

david-sawatzke:
Especially for the XGMII PHY with 64 bit width, a 32 bit datapath uses much less resources and only needs to run at 312.5 MHz, very doable on many FPGAs


enjoy-digital:
Thanks @david-sawatzke and also sorry for the delay. This looks good and remove various FIXMEs and limitations, thanks!

PR #178, opened 2024-12-16
phy/usp_gth/gty_1000basex: Set proper CPLLREFCLKSEL

FlyGoat:
As per Xilinx UG578 CPLLREFCLKSEL needs to be set to 111 when using GTGREFCLK as clock source.

Tested on VU13P.


enjoy-digital:
Thanks @FlyGoat! By curiosity, do you also tested without? IIRC if you only connect one reference clock source, CPLLREFCLKSEL can be kept at 0b001 and Vivado will automatically configure things:

image

enjoy-digital:
Hmm, sorry don't bother @FlyGoat, I could also do it and try to build your LiteX-Boards PR without and see...

FlyGoat:
Seems like this is required on my board, other wise PHY won't work, it do synthesis though.

enjoy-digital:
Thanks for the additional info, the GTGREFCLK case is maybe handled differently. This is merged, thanks!

PR #180, opened 2025-01-15
phy: rmii: use multibit io

maass-hamburg:
now that https://github.com/enjoy-digital/litex/pull/2105 is merged, we can simplify the SDR IO in the rmii phy.


enjoy-digital:
Thanks @maass-hamburg!

PR #182, opened 2025-02-03
mac: improve rx checker

maass-hamburg:
- implement padding checker, to drop packets when payload size < minimum ethernet payload size

### DIFF
--- a/liteeth/mac/padding.py
+++ b/liteeth/mac/padding.py
@@ -5,6 +5,7 @@
 # Copyright (c) 2021 David Sawatzke <d-git@sawatzke.dev>
 # Copyright (c) 2015 Sebastien Bourdeauducq <sb@m-labs.hk>
 # Copyright (c) 2018 whitequark <whitequark@whitequark.org>
+# Copyright (c) 2025 Fin Maaß <f.maass@vogl-electronic.com>
 # SPDX-License-Identifier: BSD-2-Clause
 
 import math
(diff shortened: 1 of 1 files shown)


rowanG077:
The preamble is not required by the spec to always be present on receive. When a MAC starts receiving bytes the only part that is required to be present is the SFD.

maass-hamburg:
@rowanG077 can you cite that part of the spec? I through it was mandatory, if it doesn't I wont mind adding a option to limit it to the sfd.

the check for the preamble until now depended on the selected dw. so when you had a dw of 8 it was just the sfd, but anything longer would also match the preamble. This change would only make this check behave the same on all dw.

rowanG077:
@maass-hamburg Sure...

I'm looking at the IEEE 802.3 2012 spec. It's basically sprinked all over the document but the most clear is the following:

> 4.2.6 Start frame sequence
The receiveDataValid signal is the indication to the MAC that the frame reception process should begin.
Upon reception of the sequence 10101011 following the assertion of receiveDataValid,
PhysicalSignalDecap shall begin passing successive bits to ReceiveLinkMgmt for passing to the MAC
client

And `PhysicalSignalDecap` is defined as:

> 
> ```
> procedure PhysicalSignalDecap;
> begin
> {Receive one bit at a time from physical medium until a valid sfd is detected, discard bits and return}
> end; {PhysicalSignalDecap}
> ```

Different phy links specify different things here. Some say that the full 8 bytes must be transmitted but can be corrupted. Some say full bytes can be ommited for CDR to sync.

maass-hamburg:
@rowanG077 thanks, I removed that part. One possible concern I am worried at dw bigger that 8, where the sfd is not at the end, but that would be something for another PR.

enjoy-digital:

PR #183, opened 2025-02-05
mac: Only use one CRC engine for Checker

david-sawatzke:
As we have a certain crc value for a complete valid packet, additional 0 bytes lead to a predictable new crc value. Thus, for reception, we can only use a single, wide, engine and blank out the invalid bytes. Then compare it against a precomputed expected crc value for this last_be value.

To be certain that this works, add a test to verify it.

While using the test, I noticed that 16 bit support with last_be is completely broken.

64 bits also don't work with the test case, as a new packet can arrive which the old one isn't fully processed in checker, but this *might* not be a real issue on hardware.

### DIFF
--- a/liteeth/mac/crc.py
+++ b/liteeth/mac/crc.py
@@ -159,7 +159,7 @@ class LiteEthMACCRC32Inserter(LiteXModule):
         # Parameters.
         data_width  = len(sink.data)
         ratio       = 32//data_width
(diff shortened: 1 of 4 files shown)


enjoy-digital:
Thanks @david-sawatzke  and sorry for the delay. This make sense and is merged. Thanks also for the testbench, it will be very useful for maintenance.

PR #184, opened 2025-02-07
phy: efinix: titanium: add lvds 1000base-x

maass-hamburg:
uses 1 lvds tx and 4 lvds rx of the titanium (not using the Titanium transceivers)

needs https://github.com/enjoy-digital/litex/pull/2192

for testing: https://github.com/litex-hub/litex-boards/pull/655

### DIFF
--- a/liteeth/phy/pcs_1000basex.py
+++ b/liteeth/phy/pcs_1000basex.py
@@ -15,7 +15,7 @@ from litex.gen.genlib.cdc import BusSynchronizer
 
 from litex.soc.interconnect import stream
 from litex.soc.interconnect.csr_eventmanager import *
-from litex.soc.cores.code_8b10b import K, D, Encoder, Decoder
+from litex.soc.cores.code_8b10b import K, D, Encoder, DecoderComb
 
 from liteeth.common import *
 
@@ -55,7 +55,7 @@ SGMII_100MBPS_SPEED  = 0b01
 SGMII_10MBPS_SPEED   = 0b00
 
 class PCSSGMIITimer(LiteXModule):
-    def __init__(self, speed):
+    def __init__(self, speed, valid=True):
         self.enable = Signal()
         self.done   = Signal()
 
@@ -64,13 +64,16 @@ class PCSSGMIITimer(LiteXModule):
         count = Signal(max=100)
         self.comb += self.done.eq(count == 0)
         self.sync += [
-            count.eq(count - 1),
-            If(~self.enable | self.done,
-                Case(speed, {
-                    SGMII_10MBPS_SPEED   : count.eq(99),
-                    SGMII_100MBPS_SPEED  : count.eq(9),
-                    SGMII_1000MBPS_SPEED : count.eq(0),
-                })
+            If(valid,
+                If(~self.enable | self.done,
+                    Case(speed, {
+                        SGMII_10MBPS_SPEED   : count.eq(99),
+                        SGMII_100MBPS_SPEED  : count.eq(9),
+                        SGMII_1000MBPS_SPEED : count.eq(0),
+                    })
+                ).Else(
+                    count.eq(count - 1),
+                )
             )
         ]
 
@@ -172,9 +175,10 @@ class PCSRX(LiteXModule):
         self.seen_config_reg = Signal()   # Config seen.
         self.config_reg      = Signal(16) # Config register (16-bit).
         self.sgmii_speed     = Signal(2)  # SGMII speed.
-        self.source          = source = stream.Endpoint([("data", 8)]) # Data output.
+        self.source          = source = stream.Endpoint([("data", 8), ("error", 1)]) # Data output.
+        self.input_valid     = Signal(reset=1)   # Data input valid.
 
-        self.decoder = Decoder(lsb_first=lsb_first) # 8b/10b Decoder.
+        self.decoder = DecoderComb(lsb_first=lsb_first) # 8b/10b Decoder.
 
         # # #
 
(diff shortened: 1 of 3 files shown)


liteeth/phy/titanium_1000basex.py
```diff
@@ -0,0 +1,431 @@
...
+
+
+        platform.toolchain.ifacewriter.blocks.append(block)
+        platform.toolchain.excluded_ios.append(platform.get_pin(rx_p))
+        platform.toolchain.excluded_ios.append(platform.get_pin(rx_n))
+
+class EfinixSerdesBuffer(LiteXModule):
+    def __init__(self, data_in, data_in_len, data_out, data_out_valid, allign):
+
+        self.alligner = alligner = EfinixAlligner(allign)
```
m-byte:
```suggestion
        self.aligner = aligner = EfinixAligner(align)
```

enjoy-digital (approved):
Thanks @maass-hamburg, I'm going to merge and will push some changes on top of it.

enjoy-digital:
This is now merged and reworked a bit with the following commits:
- https://github.com/enjoy-digital/liteeth/commit/209bbb2e9473df2c2f2143baf6a5896f30433747
- https://github.com/enjoy-digital/liteeth/commit/5667738884cee8a888125e37217f480ca3c2efea
- https://github.com/enjoy-digital/liteeth/commit/74c6526d80ea4beeec108a809fdd2330082161e0
- https://github.com/enjoy-digital/liteeth/commit/8aaa6cb7fa4fa1a3951e6f44f4294f16a633a203

These commits switch the code to the Synchronous 8B10B Decoder and simplify/cleanup the code a bit.

maass-hamburg:
> This is now merged and reworked a bit with the following commits:
> - https://github.com/enjoy-digital/liteeth/commit/209bbb2e9473df2c2f2143baf6a5896f30433747
> - https://github.com/enjoy-digital/liteeth/commit/5667738884cee8a888125e37217f480ca3c2efea
> - https://github.com/enjoy-digital/liteeth/commit/74c6526d80ea4beeec108a809fdd2330082161e0
> - https://github.com/enjoy-digital/liteeth/commit/8aaa6cb7fa4fa1a3951e6f44f4294f16a633a203
> 
> These commits switch the code to the Synchronous 8B10B Decoder and simplify/cleanup the code a bit.

Why was 74c6526d80ea4beeec108a809fdd2330082161e0 done? It is needed by this PR. At least for this implementation it is 100% necessary, but it depends on the used hardware.

enjoy-digital:
I would like to avoid code becoming too complex, the PCS code is expected to run at RX / TX rate and haven't found in the code where input_valid was used. If used by this implementation, can you explain the purpose and provide more info?

maass-hamburg:
> I would like to avoid code becoming too complex, the PCS code is expected to run at RX / TX rate and haven't found in the code where input_valid was used. If used by this implementation, can you explain the purpose and provide more info?

This uses it. Here we can't exactly recover the clock from the signal (clock recovery), like the other can do with their hard ip. So here we do it in logic. Normally we would get 10 bits per cycle, but to compensate for the frequency offset between the fpga and the endpoint, we can also get just 9 or 11 points per cycle. We have a buffer in between to make sure to always return 10 points or none. While none the signal will not be valid. I used the DecoderComb, so I don't had to delay that valid signal too.

enjoy-digital:
OK thanks, I'll do another review, but we should try to be as close as possible to other implementation, so if required, simplify input_valid and prefer delaying this signal instead of having a specific implementation of the Decoder.

maass-hamburg:
Also 5667738884cee8a888125e37217f480ca3c2efea should not have happened. The logic around it is time critical and we expected the result in the same cycle and not one later. This is from the logic, that detects an I2, which means we could remove it and by doing that empty our buffer by 20 points, if possible.

maass-hamburg:
> OK thanks, I'll do another review, but we should try to be as close as possible to other implementation, so if required, simplify input_valid and prefer delaying this signal instead of having a specific implementation of the Decoder.

Then we just have to delay the input valid by one cycle before it's used by the fsm

enjoy-digital:
OK thanks for the clarification. The original code wouldn't have been easy to maintain, so I'm trying to find a compromise.  We'll converge to a solution that will satisfy each other and just need to discuss things calmly and progress on each points.

maass-hamburg:
> > I would like to avoid code becoming too complex, the PCS code is expected to run at RX / TX rate and haven't found in the code where input_valid was used. If used by this implementation, can you explain the purpose and provide more info?
> 
> This uses it. Here we can't exactly recover the clock from the signal (clock recovery), like the other can do with their hard ip. So here we do it in logic. Normally we would get 10 bits per cycle, but to compensate for the frequency offset between the fpga and the endpoint, we can also get just 9 or 11 points per cycle. We have a buffer in between to make sure to always return 10 points or none. While none the signal will not be valid. I used the DecoderComb, so I don't had to delay that valid signal too.

This is still not fully fixed, 35cde07d871b6a9afc5e7b57682bd34b974c5daa is not enough, we still need the data_valid it in the fsm and the other places where it got removed.

enjoy-digital:
@maass-hamburg: Do you see a regression with this on hardware?

maass-hamburg:
> @maass-hamburg: Do you see a regression with this on hardware?

No, because it is not easy to test and I know it's still wrong.

With this we will duplicate the former symbol, if ce is not true, that will lead to a malformed packed, because one byte in in it was duplicated. Instead we should not be allowed to work with the duplicated symbol/byte.

maass-hamburg:
I would like to propose https://github.com/enjoy-digital/liteeth/pull/190

enjoy-digital:
Thanks @maass-hamburg, can you have a look at https://github.com/enjoy-digital/liteeth/commit/3b6002236864022b778dca3735a5272fe623748e? It should be similar functionally  to your PR (just a cycle of latency of difference, but should not impact things) and is simpler.

enjoy-digital:
@maass-hamburg: Does it look goog to you? If so, we could probably close https://github.com/enjoy-digital/liteeth/pull/190 and https://github.com/enjoy-digital/litex/pull/2246.

maass-hamburg:
@enjoy-digital seems working

enjoy-digital:
Thanks @maass-hamburg for the feedback.

PR #187, opened 2025-04-13
fix: UDP streamer + demo

chmousset:
The UDP streamer TX path didn't correctly set the packet size, and the bench's example didn't use the right clock domain.
The bench now works both on the TX and RX paths.

```bash
> python bench/arty.py --build --load
[...]
fpga_program
Info : ftdi: if you experience problems at higher adapter clocks, try the command "ftdi tdo_sample_edge falling"
Info : clock speed 25000 kHz
Info : JTAG tap: xc7.tap tap/device found: 0x0362d093 (mfg: 0x049 (Xilinx), part: 0x362d, ver: 0x0)
> ./bench/test_udp_streamer.py --switches --leds
# it blinks!
Switches value: 0x01
Switches value: 0x01
Switches value: 0x01
Switches value: 0x01
Switches value: 0x05
Switches value: 0x05
Switches value: 0x05
Switches value: 0x05
[...]
```

### DIFF
--- a/bench/arty.py
+++ b/bench/arty.py
@@ -5,6 +5,8 @@
 #
 # Copyright (c) 2020-2023 Florent Kermarrec <florent@enjoy-digital.fr>
 # SPDX-License-Identifier: BSD-2-Clause
+#
+# Once built, this example can be interacted with using the `bench/test_udp_streamer.py` script
 
 import os
 import argparse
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @chmousset for the PR and for pointing out the Arty UDP streamer issue.

I reviewed this against current `master`. The Stream2UDP TX packet-length part is now already covered by the newer PacketFIFO-based implementation and regression tests merged via #209, so I did not keep that older core change.

The Arty bench clock-domain part was still valid: the UDP streamer attached to the Etherbone UDP core should use Etherbone frontend clock domain so the UDP crossbar inserts the required CDC when the Ethernet core is clocked from `eth_rx`.

I added a small focused commit on current `master` for this:

- `79b908f bench: fix Arty UDP streamer clock domain`

Verified with:

- `python3 /home/florent/dev/litex/liteeth/bench/arty.py`
- `python3 -m unittest test.test_stream`

The commit credits your original PR/observation. Thanks again for the report and demo work.

chmousset:
#209 sounds better indeed :+1:

PR #194, opened 2025-09-01
phy: rmii: use rx_er if it exists

maass-hamburg:
use rx_er if it exists as it is
optional

Signed-off-by: Fin Maaß <f.maass@vogl-electronic.com>


enjoy-digital:

PR #197, opened 2026-03-26
Add jumbo frame support & fix Stream2UDPTX freeze-up

meriac:
### Why jumbo frames matter

Standard Ethernet frames are limited to a 1500-byte payload (1530 bytes including headers). For high-throughput streaming applications — such as SDR IQ sample transport — this means the CPU must process one interrupt and one DMA transaction per 1500 bytes of data. At 30.72 MSPS × 4 bytes/sample × 2 channels = 245 MB/s, that's over 160,000 packets per second, creating significant CPU overhead from interrupt handling, context switches, and per-packet protocol processing.

Jumbo frames (up to ~9000-byte payload) reduce this by ~6×: fewer packets, fewer interrupts, fewer DMA setup cycles, and better cache utilization. This directly translates to lower CPU usage and more headroom for real-time signal processing.

### What changed

Previously, `eth_mtu` was a hardcoded global constant (1530) in `liteeth/common.py`, imported via `from liteeth.common import *` by all MAC/IP/UDP modules. There was no way to override it without patching the global at runtime — a fragile approach dependent on Python import ordering.

This commit makes `eth_mtu` a proper constructor parameter threaded through the entire class hierarchy, with the original value as the default for full backward compatibility.

### Constants (`liteeth/common.py`)

- Renamed `eth_mtu` → `eth_mtu_default` (1530) to clarify it's a default, not a fixed value - and prevent accidental usage instead of the new parameter-passing approach
- Added `eth_mtu_jumboframe` (9022) as a standard jumbo frame constant
- Removed unused `buffer_depth` derived constant (would be dangerous to keep - must be calculated from the actual MTU size instead)

### Parameter threading (6 source files)

Added `eth_mtu=eth_mtu_default` parameter to the following classes, each passing it down to its children:

- **`LiteEthUDPIPCore`** / **`LiteEthIPCore`** (`core/__init__.py`) → passes to `LiteEthMAC`
- **`LiteEthMAC`** (`mac/__init__.py`) → passes to `LiteEthMACCore` and `LiteEthMACWishboneInterface`
- **`LiteEthMACCore`** (`mac/core.py`) → passes to `LiteEthMACPaddingChecker`
- **`LiteEthMACPaddingChecker`** (`mac/padding.py`) → uses `eth_mtu` for length counter signal width
- **`LiteEthMACWishboneInterface`** (`mac/wishbone.py`) → uses `eth_mtu` for SRAM depth calculation and Wishbone address decoding; passes to `LiteEthMACSRAM`
- **`LiteEthMACSRAM`** / **`LiteEthMACSRAMWriter`** (`mac/sram.py`) → uses `eth_mtu` for the packet truncation threshold (`If(length >= self.eth_mtu, DISCARD-REMAINING)`)

### Tests (6 test files)

All test files now run at both `eth_mtu_default` (1530) and `eth_mtu_jumboframe` (9022) using `unittest.subTest`:

- `test_arp.py`, `test_icmp.py`, `test_ip.py`, `test_udp.py`, `test_etherbone.py` — DUT accepts `eth_mtu` parameter, test loops over both MTU values
- `test_mac_wishbone.py` — additionally replaces hardcoded SRAM slot offsets (`0x200`) with dynamically computed values based on `eth_mtu` and `dw`, since the Wishbone address decoder requires power-of-2 aligned slot boundaries that scale with MTU

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ nosetests.xml
 coverage.xml
 *,cover
 .hypothesis/
+sim.vcd
 
 # Translations
 *.mo
--- a/liteeth/common.py
+++ b/liteeth/common.py
@@ -18,12 +18,12 @@ from litex.soc.interconnect.packet import Header, HeaderField
 
 # Ethernet Constants -------------------------------------------------------------------------------
 
-eth_mtu              = 1530
+eth_mtu_default      = 1530
+eth_mtu_jumboframe   = 9022
 eth_min_frame_length = 64
 eth_fcs_length       = 4
 eth_interpacket_gap  = 12
 eth_preamble         = 0xd555555555555555
-buffer_depth         = 2**log2_int(eth_mtu, need_pow2=False)
 
 ethernet_type_ip     = 0x800
 ethernet_type_arp    = 0x806
--- a/liteeth/core/__init__.py
+++ b/liteeth/core/__init__.py
@@ -27,6 +27,7 @@ class LiteEthIPCore(LiteXModule):
         rx_cdc_buffered   = True,
         interface         = "crossbar",
         endianness        = "big",
+        eth_mtu           = eth_mtu_default,
     ):
         # Parameters.
         # -----------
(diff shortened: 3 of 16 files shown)


enjoy-digital:
Thanks @meriac! This looks good.

PR #203, opened 2026-06-08
phy/agilex_rgmii: reworks RX data path with correct primitives, catch align/unalign stream and allows disabling PHY reset

trabucayre:
This PR has 3 goals:
- Fix primitives for TX data path. This is based on official core produces by qsys. This also fixes timing not met issues
- Sometime (mainly at start time) data are not correctly aligned. An FSM is added to check according to ctrl signal if ddr input is correctly aligned.
- In some situation/boards (like the terasic Atum A3 nano) the PHY is really slow to be ready after reset. An optional parameter is added to bypass the reset PHY. In this situation the signal is fixed.

Tested with the terasic Atum A3 Nano to boot demo and linux

### DIFF
--- a/liteeth/phy/agilex_rgmii.py
+++ b/liteeth/phy/agilex_rgmii.py
@@ -37,23 +37,28 @@ def io_obuf(i, o):
         )
     ]
 
-def io_ibuf(i, o):
+def ddio_in(clk, i, o1, o2):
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @trabucayre, this is merged.

PR #204, opened 2026-06-30
Add packet fifo based backpressure

inochisa:
Add packet fifo based backpressure module for `LiteEthMACSRAM`

This backpressure module includes some common packet assembly/disassembly logic for the sram module. I hope it could be a good start point for the DMA based ethernet.

For RX logic, most common logic is moved into the new packet class so the SRAM writer class reduced significantly. However, the TX logic movement is not very good as still many logic is SRAM specifically.

This module also introduce a small resource waste as the packet module must buffer at least one packet, otherwise the module will fail. It may be improved with something bypass in the future.

This PR is tested on VexiiRiscv@200Mhz (with mainline kernel 7.2-rc1):
```log
[root@ku5p-test ~]# iperf3 -c 192.168.1.151                                                                                                                                                                                            
Connecting to host 192.168.1.151, port 5201
[  5] local 192.168.1.182 port 36580 connected to 192.168.1.151 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.00   sec  11.4 MBytes  95.1 Mbits/sec    0    132 KBytes       
[  5]   1.00-2.00   sec  2.62 MBytes  22.1 Mbits/sec    0    132 KBytes       
[  5]   2.00-3.00   sec  11.1 MBytes  93.3 Mbits/sec    0    132 KBytes       
[  5]   3.00-4.00   sec  6.62 MBytes  55.6 Mbits/sec    0    132 KBytes       
[  5]   4.00-5.36   sec  0.00 Bytes  0.00 bits/sec    0    132 KBytes       
[  5]   5.36-7.41   sec  0.00 Bytes  0.00 bits/sec    0    132 KBytes       
[  5]   7.41-7.43   sec   128 KBytes  47.8 Mbits/sec    0    132 KBytes       
[  5]   7.43-8.00   sec  2.50 MBytes  37.0 Mbits/sec    0    132 KBytes       
[  5]   8.00-9.16   sec  0.00 Bytes  0.00 bits/sec    0    132 KBytes       
[  5]   9.16-10.04  sec  4.00 MBytes  38.1 Mbits/sec    0    132 KBytes       
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.04  sec   115 MBytes  95.9 Mbits/sec    0            sender
[  5]   0.00-10.07  sec   115 MBytes  95.6 Mbits/sec                  receiver

[...]

PR #212, opened 2026-07-01
mac: add optional MAC-level packet FIFOs without default FIFO cost

enjoy-digital:
This is a fresh, focused follow-up to #204, based on original packet FIFO/backpressure work from Inochi Amaoto. The original contributor branch cannot be updated directly (`maintainer_can_modify=false`), so this keeps the useful pieces on an enjoy-digital branch while preserving credit in the original commits.

The latest version also follows the review direction from @inochisa: the optional complete-packet FIFOs are owned by `LiteEthMAC` at the Wishbone/Hybrid MAC boundary, not by the low-level SRAM packet reader/writer helpers. The SRAM helpers are direct packet frontends again; they keep the slot/status/drop/memory policy, while FIFO elasticity is a MAC-level configuration choice.

## Improvements vs current master

- The SRAM RX/TX packet boundary logic is factored into explicit `LiteEthMACPacketWriter` / `LiteEthMACPacketReader` frontends, making length, `last_be`, MTU/error, timestamp, completion, drop, and disabled-path handling testable in isolation.
- The default Wishbone/Hybrid SRAM path keeps `rx_fifo_depth=0` / `tx_fifo_depth=0`, so current users do not pay for complete-packet FIFO payload memories.
- With the default `rx_fifo_depth=0`, RX keeps the master-compatible behavior: if no status FIFO slot is available, the packet is consumed/dropped and reported as a drop instead of propagating non-actionable backpressure into the Ethernet RX stream without any internal packet storage.
- With `rx_fifo_depth > 0`, `LiteEthMAC` inserts an RX `PacketFIFO` before the SRAM writer. The SRAM writer can then hold when no status slot is available, and the MAC-level FIFO provides the actual complete-packet elasticity/backpressure.
- With `tx_fifo_depth > 0`, `LiteEthMAC` inserts a TX `PacketFIFO` between the SRAM reader and MAC core.
- The public configuration names use the shorter direction+FIFO style already used by LiteEth stream frontends: `rx_fifo_depth` / `tx_fifo_depth`.
- Packet frontend and wrapper regressions now cover 8/16/32/64-bit datapaths, good packets, oversize/error drops, disabled/drop behavior, disabled/hold behavior, reader `last_be`/`done`, timestamp capture, direct zero-depth mode, and the MAC-boundary FIFO wrapper.

## CSR/Linux Compatibility

I checked the default Wishbone MAC configuration used by the existing tests against `origin/master` (`79b908f`) and this PR head (`bce752c`). The CSR ABI is unchanged.

`LiteEthMAC(..., interface="wishbone", dw=32, with_preamble_crc=True)` reports the same `get_csrs()` list on both revisions:

```text
0  sram_writer_slot                 CSRStatus   1
1  sram_writer_length               CSRStatus   11
2  sram_writer_errors               CSRStatus   32
3  sram_writer_ev_status            CSRStatus   1
4  sram_writer_ev_pending           CSRStatus   1
5  sram_writer_ev_enable            CSRStorage  1
6  sram_reader_start                CSR         1
7  sram_reader_ready                CSRStatus   1
8  sram_reader_level                CSRStatus   2
9  sram_reader_slot                 CSRStorage  1
10 sram_reader_length               CSRStorage  11
11 sram_reader_ev_status            CSRStatus   1
12 sram_reader_ev_pending           CSRStatus   1
13 sram_reader_ev_enable            CSRStorage  1
14 preamble_crc                     CSRStatus   1
15 rx_datapath_preamble_errors      CSRStatus   32
16 rx_datapath_crc_errors           CSRStatus   32
```

The constants are also unchanged:

```text
rx_slots  = 2
tx_slots  = 2
slot_size = 2048
```

No new CSR or EventManager is added by the packet frontends or by the optional packet FIFO wrapper. The new packet FIFO controls are Python constructor parameters only, so the default Linux/software-visible CSR layout stays compatible.

## Credit

- `fd9db31 core: add packet fifo support to SRAM interface` authored by Inochi Amaoto;
- `f72302e test: add test for packet reader/writer` authored by Inochi Amaoto;
- `bce752c mac: move packet FIFOs to MAC boundary` follows @inochisa's review feedback by moving the optional FIFO policy out of the SRAM packet helpers and into `LiteEthMAC`.

## Arty Implementation Check

`digilent_arty`, 32-bit Wishbone MAC harness, Vivado 2024.1. The harness touches both RX/TX SRAM slots so the numbers are comparable.

| Variant | LUTs | FFs | LUT RAM | BRAM tiles | WNS |
|---|---:|---:|---:|---:|---:|
| master (`79b908f`) | 392 | 460 | 6 | 5 | 1.215 ns |
| this PR (`bce752c`), default `rx_fifo_depth=0`, `tx_fifo_depth=0` | 435 | 463 | 6 | 5 | 1.215 ns |
| this PR (`bce752c`), `rx_fifo_depth=1`, `tx_fifo_depth=1` | 1310 | 539 | 606 | 5 | 1.236 ns |

So the default configuration adds a small amount of packet-control logic (`+43` LUTs, `+3` FFs in this harness) but no extra LUT RAM or BRAM, and the default generated hierarchy does not instantiate `LiteEthMACPacketFIFOs`. The large LUT RAM increase appears only when complete-packet FIFOs are explicitly enabled, as expected from the two MTU-sized `PacketFIFO` payload memories.

## Tested

- `python3 -m unittest test.test_mac_packet`
- `python3 -m unittest test.test_mac_wishbone`
- `python3 -m unittest test.test_mac_core`
- `python3 -m py_compile liteeth/mac/packet.py liteeth/mac/sram.py liteeth/mac/wishbone.py liteeth/mac/__init__.py test/test_mac_packet.py`
- `git diff --check`
- Default Wishbone MAC CSR list compared against `origin/master` (`79b908f`).
- Digilent Arty Vivado builds/bitstreams completed and met timing for master, this PR default, and this PR with `rx_fifo_depth=1`, `tx_fifo_depth=1`.

Draft for review of the API/default policy before merging.

### DIFF
--- a/liteeth/mac/packet.py
+++ b/liteeth/mac/packet.py
@@ -18,7 +18,8 @@ from liteeth.common import *
 # MAC Packet Writer Frontend -----------------------------------------------------------------------
 
 class LiteEthMACPacketWriter(LiteXModule):
-    def __init__(self, dw, depth, eth_mtu=eth_mtu_default, fifo_depth=1, timestamp=None):
+    def __init__(self, dw, depth, eth_mtu=eth_mtu_default, fifo_depth=1, timestamp=None,
+        drop_when_disabled=False):
         # Endpoint / Signals.
         self.sink   = sink   = stream.Endpoint(eth_phy_description(dw))
         self.source = source = stream.Endpoint(eth_packet_description(dw))
@@ -84,8 +85,17 @@ class LiteEthMACPacketWriter(LiteXModule):
         # FSM.
         self.fsm = fsm = FSM(reset_state="IDLE")
         fsm.act("IDLE",
-            If(packet_source.valid & self.enable,
-                NextState("WRITE")
+            If(packet_source.valid,
+                If(self.enable,
+                    NextState("WRITE")
+                ).Elif(drop_when_disabled,
+                    packet_source.ready.eq(1),
+                    If(packet_source.last,
+                        NextState("DISCARD")
+                    ).Else(
+                        NextState("DISCARD-ALL")
+                    )
+                )
             )
         )
         fsm.act("WRITE",
--- a/liteeth/mac/packet.py
+++ b/liteeth/mac/packet.py
@@ -80,12 +80,8 @@ class LiteEthMACPacketWriter(LiteXModule):
         # FSM.
         self.fsm = fsm = FSM(reset_state="IDLE")
         fsm.act("IDLE",
-            If(packet_fifo.source.valid,
-                If(self.enable,
-                    NextState("WRITE")
-                ).Else(
-                    NextState("DISCARD-ALL")
-                )
+            If(packet_fifo.source.valid & self.enable,
+                NextState("WRITE")
             )
         )
         fsm.act("WRITE",
--- a/liteeth/mac/packet.py
+++ b/liteeth/mac/packet.py
@@ -1,0 +1,225 @@
+#
+# This file is part of LiteEth.
+#
+# Copyright (c) 2015-2021 Florent Kermarrec <florent@enjoy-digital.fr>
+# Copyright (c) 2015-2018 Sebastien Bourdeauducq <sb@m-labs.hk>
+# Copyright (c) 2021 Leon Schuermann <leon@is.currently.online>
+# Copyright (c) 2017 whitequark <whitequark@whitequark.org>
+# SPDX-License-Identifier: BSD-2-Clause
+
+import math
+
+from litex.gen import *
+
+from litex.soc.interconnect.packet import PacketFIFO
+
+from liteeth.common import *
+
+# MAC Packet Writer Frontend -----------------------------------------------------------------------
+
+class LiteEthMACPacketWriter(LiteXModule):
+    def __init__(self, dw, depth, eth_mtu=eth_mtu_default, fifo_depth=1, timestamp=None):
+        # Endpoint / Signals.
+        self.sink   = sink   = stream.Endpoint(eth_phy_description(dw))
+        self.source = source = stream.Endpoint(eth_packet_description(dw))
+
+        self.enable  = Signal(reset=1)
+        self.done    = Signal()
+        self.drop    = Signal()
+        self.error   = Signal()
+        self.offset  = Signal(bits_for(depth*dw//8))
+        self.length  = Signal(bits_for(depth*dw//8))
+
+        # Parameters Check / Compute.
+        assert dw in [8, 16, 32, 64]
+        assert depth > 0
+        assert fifo_depth > 0
+        if timestamp is not None:
+            timestampbits  = len(timestamp)
+            self.timestamp = Signal(timestampbits)
+
+        # # #
+
+        length  = Signal.like(self.length)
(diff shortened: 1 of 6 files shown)


liteeth/mac/sram.py
```diff
@@ -48,7 +49,13 @@ def __init__(self, dw, depth, nslots=2, endianness="big", timestamp=None, eth_mt
...
 
         # Packet frontend.
-        self.packet = packet = LiteEthMACPacketWriter(dw, depth, eth_mtu=eth_mtu, timestamp=timestamp)
+        self.packet = packet = LiteEthMACPacketWriter(
+            dw         = dw,
+            depth      = depth,
+            eth_mtu    = eth_mtu,
+            fifo_depth = packet_fifo_depth,
+            timestamp  = timestamp,
+        )
```
inochisa, line 58: I think we should make an agreement: it is better to move this into LiteEthMAC. This is allowed as we do not have a DMA engine now.

liteeth/mac/packet.py
```diff
@@ -204,18 +221,31 @@ def __init__(self, dw, depth, fifo_depth=1, timestamp=None):
...
                 NextState("IDLE"),
-            )
+            )]
+        else:
+            end_statements = [
+                self.done.eq(1),
+                NextState("IDLE"),
+            ]
+        fsm.act("END",
+            *end_statements,
```
inochisa: Emmm, I wonder whether there is a better way to do this, this is something why I do not add the zero-depth support....

inochisa (approved):

inochisa:
@enjoy-digital Anyway, I think it is good enough.

This only thing I am concerned about is the usage of LiteEthMACPacketWriter, other things is a good improvement. If you are OK, feel free to close my PR and merge yours.

liteeth/mac/sram.py
inochisa: And this is why I did not move it to the `LiteEthMAC`.

enjoy-digital:
Thanks @inochisa for the quick feedback, I'll review it tomorrow.

enjoy-digital:
Thanks @inochisa, I reworked this in the direction you suggested in bce752c.

The optional `PacketFIFO` storage is now owned by `LiteEthMAC` at the Wishbone/Hybrid MAC boundary. `LiteEthMACSRAMWriter` / `LiteEthMACSRAMReader` are direct packet helpers again and only keep the SRAM slot/status/drop policy.

For the default `rx_fifo_depth=0` path, RX still consumes/drops when no status slot is available, which avoids exposing non-actionable RX backpressure without internal storage. When `rx_fifo_depth > 0`, the SRAM writer holds instead and the MAC-level `PacketFIFO` provides the actual complete-packet buffering/backpressure.

I also added focused MAC-boundary FIFO tests, checked that the default CSR ABI is unchanged against `origin/master`, and reran the packet/wishbone/core tests plus Digilent Arty implementation checks.

enjoy-digital:
@inochisa: I merged this since easier, but next time we could keep your PR if you allow me to push modifications to it.

PR #216, opened 2026-08-28
phy: add experimental Artix-7 5000BASE-X support

enjoy-digital:
## Summary

- add a 32-bit, four-symbol Clause-36/37 PCS for an experimental 5Gbps link
- configure the Artix-7 GTPE2 40-bit fabric interface with 312.5MHz `USRCLK` and 156.25MHz `USRCLK2`/PCS clocking
- expose optional loopback, PRBS, alignment, and reset diagnostics
- integrate the PHY with the LiteEth generator and add simulation/unit coverage

This is intentionally experimental `5000BASE-X`: 6.25Gbaud NRZ with 8b/10b encoding. It is not IEEE 5GBASE-KR/R or 10GBASE-R.

## Validation

- full LiteEth suite: 240 tests and 139 subtests passed
- post-rebase/style focused suite: 30 tests and 8 subtests passed
- Vivado 2024.1 M2SDR integration: WNS +0.037ns, WHS +0.022ns
- Artix-7 near-PMA PRBS31: approximately 63.1 billion bits with zero error indications; a deliberate injected error was observed
- normal-traffic near-PMA loopback: byte alignment, PCS link, and Clause-37 negotiation completed

External module compatibility still requires a module with an explicit 6.25Gbaud host-side mode.


storborg:
I got this working nearly as-is on the Acorn with [this H!Fiber ASF-10G-T80 transceiver module](https://www.amazon.com/dp/B09MJYPFMM), which uses a Broadcom BCM84891 PHY. A few notes:
* The module had to be aggressively persuaded to switch to that PHY mode. It implements RollBall, but defaults to a mode which is 10GBASE-R only on the host side interface. I'll explore this more and work on getting it integrated into the BIOS.
* I had to disable the autonegotiation FSM in the PCS and replace it with hardcoded signals (`align = 1, link_up = ~checker_error`). I think this is a quirk of the module, but I ran into similar issues experimenting with other PHYs: it seems like a lot of modules can't correctly handle AN for 2500BASE-X, 5000BASE-X, 5GBASE-R, etc. I think this merits a flag which disables AN in the PCS (and subsequent PHY wrapper) instantiation.

This transceiver seems to have both the widest range of host-side interfaces *and* the lowest power consumption, so it's the winner so far. But in the US it's quite a bit more expensive than the Marvell/Aquantia AQR113C based modules. Both of these claim to properly support PTP, but I haven't tested that yet.

PR #151, opened 2023-11-12
core: Add buffers to IP and ICMP

rowanG077:
This significantly improves timing on ECP5. This allows ECP5 to once again reach 50Mhz with ease.

It does burn a few FFs and LUTs though On a UDP echo design on a 25k ECP5 we go from

```
Info:             TRELLIS_FF:  2334/24288     9%
Info:           TRELLIS_COMB:  5432/24288    22%
Info:           TRELLIS_RAMW:   136/ 3036     4%
```

to

```
Info:             TRELLIS_FF:  2570/24288    10%
Info:           TRELLIS_COMB:  5752/24288    23%
Info:           TRELLIS_RAMW:   136/ 3036     4%
```

So if you want to make this configurable @enjoy-digital let me know

PR #217, opened 2026-08-28
phy: add Artix-7 5GBASE-R support

enjoy-digital:
## Summary

- add a reusable 64b/66b BASE-R PCS around LiteEth's 64-bit XGMII frontend
- add an Artix-7 GTP `A7_5000BASER` PHY at 5.15625 Gbaud using the native 32-bit gearbox
- model the native-gearbox clock cadence and pipeline the XGMII-to-encoder boundary for timing closure
- retain the existing 1000BASE-X and 2500BASE-X behavior

The PCS RTL is derived from Alex Forencich's MIT-licensed
[`verilog-ethernet`](https://github.com/alexforencich/verilog-ethernet/tree/baac5f8d811d43853d59d69957975ead8bbed088/rtl).
The LFSR instances explicitly select the loop implementation to keep Vivado
synthesis tractable.

## Validation

- `pytest -q`: 232 tests and 131 subtests passed on current `master`
- focused Artix-7/XGMII tests: 3 tests passed
- direct Verilator PCS loopback: block lock acquired, BER stayed clear, and decoded XGMII idles matched
- Vivado 2024.1 LiteX-M2SDR baseboard build with the receive LiteScope probe generated a bitstream
- all new 5GBASE-R clock groups meet setup and hold timing; the slowest new crossing has +0.019 ns hold and +0.063 ns setup slack

The complete M2SDR image has an unrelated -0.003 ns setup margin in its
existing 245.76 MHz RFIC clock group. The BASE-R intra-clock groups have
+6.730 ns RX and +9.800 ns TX setup slack.


storborg:
@enjoy-digital, have you validated this with a real link partner yet? If so, could you share info about what devices / transceivers you have used?

enjoy-digital:
@storborg: Not yet, only in loopback for now.

storborg:
I've been hunting for test hardware here, and I have learned some interesting new-to-me things about SFP+ transceivers and 802.3ae PHYs in general.

There are plenty of copper SFP+ transceivers that can run at 1/2.5/5/10GBASE-T rates. I had assumed that, when these slowed down to 5Gb, they would use 5GBASE-R on the SFP+ side. This was wrong: they continue to use 10GBASE-R, regardless of the rate they are speaking over the RJ-45 port. I'm not sure if this is universal, but it was the case for all six of the modules I have (labelled as three different vendors). These don't support autonegotiation on the SFP+ side, so they're effectively stuck at 10GBASE-R.

I looked around exhaustively for off-the-shelf 5GBASE-R devices and found almost nothing. It seems to just not exist in the real world. The one exception is a [media converter from KTI Networks](https://www.ktinet.com.tw/product/ktgc-200/) that implies that it can do 5GBASE-R, but it's a little pricey and not readily available in the USA. This isn't particularly encouraging for making an Acorn/M2SDR on a Baseboard Mini interface directly to an optical switch, but maybe we can still get a 5GBASE-T link on it...

Along the way, I did confirm that several Aquantia/Marvell PHYs commonly used in SFP+ transceivers theoretically support 5GBASE-R upstream. I set up the robot (used Claude Opus here) to bruteforce I2C access schemes on every one of my 10GBASE-T modules. Lo and behold, on exactly one of them, it managed to figure out the write sequence of I2C transactions to unlock the RollBall MDIO-over-I2C protocol. This allows direct writes to the PHY configuration registers, and the module can now speak 5GBASE-R upstream instead of 10GBASE-R.

Here's a PoC script that does the mode conversion:
https://gist.github.com/storborg/d38fa2664e6cf2e83344d93b4818e381

This is the exact module that worked:
https://www.amazon.com/dp/B07P39G4XJ

No luck on enabling any register access on any other modules, even others labeled with the same SFP-10G-T part number and advertised to contain the same PHY. I suspect this comes down to somewhat simpler MCU code: they didn't bother implementing RollBall at all.

Unfortunately, I haven't been able to determine any mechanism for making this change persistent, so it has to be done on every module power-on. This might involve writing new firmware to the module's MCU, which may be locked. Here are some clues on a related but different module:
https://forums.servethehome.com/index.php?threads/teardown-of-the-adop-10gms-30m-t-10gbase-t-adapter.29864/

enjoy-digital:
@storborg: Awesome, thanks for sharing! I did a quick search when doing this PR and was going to spend more time on this... :)

storborg:
Found this as well, putting it here for reference:
https://github.com/alexforencich/verilog-ethernet/discussions/189#discussioncomment-13284194

I suspect Wiitek module I used is just a white labeled variant of this, given how close the feature list text is:
https://www.hilinktech.com/copper-transceivers/10gbase-t-sfp.html

storborg:
I've been trying to make this branch work, using the [M.2 SDR target](https://github.com/enjoy-digital/litex_m2sdr/tree/feature/baseboard-5000baser) as a reference. So far I've been unsuccessful at even getting `block_lock` set at all on a PMA loopback.

One thing that looks suspicious: [UG482 Table 4-45](https://docs.amd.com/v/u/en-US/ug482_7Series_GTP_Transceivers) says that in 32-bit data path mode, RXUSRCLK2 should be half the frequency of RXUSRCLK. But here they are the same:
https://github.com/enjoy-digital/liteeth/blob/feature/a7-5000baser/liteeth/phy/a7_1000basex.py#L682-L683

I'm adapting to a CLE-215+ on the Acorn Baseboard Mini, so it's a less appropriate clock tree getting the reference from fabric via GTGREFCLK. This is warned against in the GTP guide, not sure how seriously to take those warnings.

enjoy-digital:
Thanks @storborg for you SFP modules findings and test of the branch. Since this PR was indeed just a preparation of the work without the right SFP module yet, there are probably a few things to fix. I'll try to order the one of the working module to be able to continue the work.

enjoy-digital:
Some notes:

# 5GBASE-R over copper SFP+ modules / Marvell 88X3310 notes

The main difficulty is not the LiteEth 5GBASE-R PCS/PMA itself, but finding a copper SFP+ module that can actually expose **5GBASE-R on the SFP+ host side**.

Most 1/2.5/5/10GBASE-T copper SFP+ modules keep the host SerDes at **10.3125 Gb/s / 10GBASE-R** and rate-match internally when the RJ45 side negotiates 5GBASE-T.

The interesting exception is modules based on Marvell Alaska PHYs such as the **88X3310**, where the PHY can be configured so that the host interface follows the negotiated copper rate:

| Copper side | Host side |
|---|---|
| 1GBASE-T | SGMII / 1000BASE-X depending on configuration |
| 2.5GBASE-T | 2500BASE-X |
| 5GBASE-T | **5GBASE-R** |
| 10GBASE-T | 10GBASE-R |

Linux also describes this 88X3310 mode as a MACTYPE where the MAC-side interface follows the copper rate, versus the usual 10GBASE-R rate-matching mode.

References:

- Linux `marvell10g.c`: https://github.com/torvalds/linux/blob/master/drivers/net/phy/marvell10g.c
- Linux PHY-mode discussion: https://www.spinics.net/lists/devicetree/msg413869.html

## Confirmed working module

The currently confirmed module is the Wiitek `SFP-10G-T` sold on Amazon:

- Amazon: https://www.amazon.com/dp/B07P39G4XJ
- Manufacturer shown by Amazon: **Wiitek Technology**
- Model: **SFP-10G-T**

This is the module used by @storborg in the PR discussion.

It exposes the Marvell PHY registers through the **RollBall MDIO-over-I2C protocol**.

PoC:

https://gist.github.com/storborg/d38fa2664e6cf2e83344d93b4818e381

### RollBall access

The PHY is not accessed through dedicated MDIO/MDC pins. The SFP only needs its normal I2C interface.

The working module uses:

- I2C address: `0x51`
- unlock password: `0xffffffff`
- vendor offset: `123`
- SFP page: `3`
- MDIO command/data window around `0x80..0x85`

This tunnels **Clause 45 MDIO** over the SFP I2C bus.

For the 88X3310:

```text
PHY ID:              0x002b09a0
MMD:                 31
Port Control reg:    0xf001
MACTYPE mask:        0x7

MACTYPE 6:
    10GBASE-R rate matching
    host SerDes stays at 10.3125 Gb/s

MACTYPE 4:
    host SerDes follows copper speed
    -> 5GBASE-T produces 5GBASE-R
```

The PoC effectively does:

```text
read  MMD31:0xf001
change MACTYPE[2:0] from 6 -> 4
write MMD31:0xf001
read back / verify
set SWRST (bit 15)
```

After the software reset the copper side renegotiates and the host SerDes follows the negotiated rate.

## HiLink module

The HiLink copper SFP+ documentation is especially interesting because it explicitly advertises:

- `XFI / 5GBASE-R / 2500BASE-X / SGMII`
- access to the physical-layer IC through the 2-wire serial bus

https://www.hilinktech.com/copper-transceivers/10gbase-t-sfp.html

This looks very close to the behavior required for LiteEth #217 and may be the original/reference design behind some white-labelled modules.

## Possible alternative: 10Gtek ASF-10G-T-3310

A promising alternative is:

**10Gtek ASF-10G-T-3310**

10Gtek explicitly documents this SKU as using the **Marvell 88X3310**:

https://store.10gtek.com/10gbase-t-10g-sfp-30m-copper-rj-45-cat-6a-cat-7-transceiver-module/p-5226

Important caveat:

> Having an 88X3310 inside the module is not sufficient.

The PR experiments already show that modules with apparently similar hardware / labels do not necessarily expose the RollBall MDIO window. This depends on the SFP module MCU firmware.

So the 10Gtek module is a good candidate to test, but **RollBall access is not yet confirmed**.

When buying one, I would specifically look for:

```text
ASF-10G-T-3310
```

and avoid assuming that similarly named newer modules use the same PHY or MCU firmware.

## Configuration persistence

For the known working Wiitek/RollBall module, the MACTYPE change is **not persistent**.

The PoC explicitly observes that a module power cycle restores the factory configuration.

So with an unmodified module we currently need to configure the PHY:

- after module power-up,
- after hot-plug,
- and probably after any reset that resets the internal PHY.

This does **not** require continuously driving MDIO/MDC.

It is just a short initialization sequence over the normal SFP I2C bus.

A reasonable LiteX implementation would therefore be:

```text
SFP inserted / board boot
        |
        v
probe I2C 0x51
        |
        v
try RollBall unlock
        |
        v
read Clause-45 PHY ID
        |
        v
if 88X3310:
    read MMD31:0xf001
    if MACTYPE != 4:
        write MACTYPE = 4
        verify
        issue SWRST
```

Once configured, no further management traffic is required for normal packet operation.

## Could the configuration be stored permanently?

Probably **technically yes**, but not through the normal runtime register interface.

The 88X3310 itself supports booting firmware/configuration from non-volatile SPI flash. Linux development around this PHY explicitly distinguishes:

- firmware loaded into volatile PHY RAM over MDIO, and
- firmware stored in non-volatile SPI flash.

Reference:

https://www.spinics.net/lists/netdev/msg645238.html

So there is non-volatile storage somewhere in a normal 88X3310 design.

However, for an SFP+ module there are usually two firmware/configuration layers:

```text
        SFP I2C
           |
           v
      module MCU
           |
      RollBall / MDIO
           |
           v
      Marvell 88X3310
           |
      optional SPI flash
```

Making MACTYPE=4 permanent could therefore require one of:

1. changing the 88X3310 SPI configuration/firmware;
2. modifying the module MCU firmware so it programs MACTYPE=4 at boot;
3. modifying whatever factory configuration image the module loads.

Those mechanisms are vendor-specific, undocumented in these modules, and potentially brick the SFP.

For LiteEth I would therefore **not depend on persistent modification of the module**.

## Suggested LiteEth/LiteX integration

I would keep the module-specific configuration outside the BASE-R PCS itself.

Something like:

```text
liteeth/phy/
    a7_5000baser.py       # generic FPGA 5GBASE-R PHY

LiteX board/target firmware:
    SFP I2C
      + optional RollBall helper
      + 88X3310 initialization
```

or possibly a small reusable LiteX-side helper for RollBall-compatible modules.

The gateware requirement is only an I2C master, which is already useful for normal SFP EEPROM access.

The actual initialization is small enough to do in BIOS/software rather than building a dedicated MDIO state machine in gateware.

That also lets us:

- identify the PHY before touching anything;
- gracefully reject non-RollBall modules;
- inspect the existing MACTYPE and no-op if already correct;
- support additional Marvell PHY variants later.

## Practical next step

The best next experiment seems to be ordering one or more **10Gtek ASF-10G-T-3310** modules and checking whether they implement the same RollBall window:

```text
I2C 0x51
page 3
unlock 0xffffffff
read Clause-45 PHY ID
```

If it returns an 88X3310 and exposes MMD31 register `0xf001`, it should be straightforward to test it with the existing PoC and LiteEth 5GBASE-R branch.

If it does not expose RollBall, the presence of the 88X3310 alone unfortunately does not help without opening/modifying the module.

enjoy-digital:
@storborg: With LiteX, it's already possible to do a I2C init directly from gateware, ex as done for Video component init: https://github.com/litex-hub/litex-boards/blob/master/litex_boards/targets/lambdaconcept_ecpix5.py#L215-L235, we could also probably use this for 5GBASE-R init.

storborg:
I actually tried [that 10GTek module](https://store.10gtek.com/10gbase-t-10g-sfp-30m-copper-rj-45-cat-6a-cat-7-transceiver-module/p-5226) and it didn't expose RollBall. The other failure was the [FS SFP-10G-T-30](https://www.fs.com/products/66613.html?now_cid=63). That said, I wonder if these non-RollBall modules may actually be more suited to permanent modification, by just modifying the SPI flash directly via test points.

Another promising module is the [Mikrotik S+RJ10](https://mikrotik.com/product/s_rj10). This is the only module I've seen advertised as using an 88X3310P, which is the variant of the 88X3310 that supports PTP.

Other notes:
* Toggling `mod_disable` seems to reliably reset the PHY on these.
* There are useful loopback modes in the PHY as well.

I agree that this I2C tooling is best kept out of the 5GBASE-R modules. I think the BIOS-based solution is the best approach. Regarding doing it from gateware, I played around with using a microcoded state machine controller to configure this module at power-on. It does work, but I think it's much more brittle than C BIOS code--the microcode needs to be pretty huge in order to support as many modules as possible, and needs polling states to be robust to interaction timing. I wouldn't want to have to maintain it long-term.

storborg:
Ok, with many minor fixes, I've been able to get this working on the Acorn Baseboard Mini + CLE-215, the AC701, and the KC705. The biggest debugging obstacles were:

1. A [hardware issue on the Baseboard Mini](https://github.com/enjoy-digital/litex-acorn-baseboard/issues/22) caused the 5GBASE-R PHY to reset randomly.
2. The MAC and crossbars suffer from the same issue I alluded to in the [10GBASE-R thread](https://github.com/enjoy-digital/liteeth/pull/218#issue-5279731626) and haven't posted a proper fix for yet: packets get mangled by individually dropped words because the internal logic can't keep up with the line rate.
3. The Acorn clocking is a little dodgey with the fabric-routed GTGREFCLK reference, though it does work with everything else fixed. On the AC701, I routed the clock through the Si5234 jitter cleaner, but then the GTP required a reset *after* the initialization of the Si5234.

My working tree is a total mess of probes and hacks right now, so I'm working on minimizing the diff back down, then I can submit a reference PR against this branch with the fixes. However, what I would propose is a cleaned-up PR combining:
- 7-series PMAs from this branch: 5GBASE-R on GTP, 5GBASE-R on GTX, 10GBASE-R on GTX
- UltraScale+ PMAs from the other branch: 10GBASE-R on GTH, 5GBASE-R on GTY, 10GBASE-R on GTY, and maybe 25GBASE-R on GTY
- LiteX-native PCS from the other branch

Separately, I have some other fixes that I think are important in order to do anything useful with these PHYs. The demo testbenches depend on all of these so maybe I'll try to get PRs sorted for these first.
1. RollBall PHY setup tools in the BIOS, controlled by config flags. This is pretty low impact.
2. Opt-in packet-dropping FIFO semantics on IP and UDP crossbar port creation.
3. Additional pipelining in the MAC, IP Rx path, and crossbar dispatcher. This isn't finished/fully working yet. Like you mentioned in #151, it includes registering the backpressure path. This makes it possible to close timing at 390.625 MHz (25Gb, dw 64) on US+ speed grade 2, but I think it would also be helpful on other platforms. So far I'm around 1-2 cycles of additional latency on Tx and 4-7 cycles on Rx. If you're concerned about keeping the MAC as low latency as possible I could probably set those up to be opt-in.

Let me know if there's anything I can do to help make this easier to review / merge, or if you'd like to discuss via email or chat. This turned into a much larger project than I anticipated!

storborg:
I bought another one of the WiiTek transceivers from the [exact same Amazon product listing as before](https://www.amazon.com/dp/B07P39G4XJ) and the new module doesn't work: it doesn't seem to expose any PHY registers over either the RollBall protocol or the SFP Enhanced Digital Diagnostic Interface.

I also bought a Mikrotik S+RJ10 and got similar results.

I think it will be frustrating to reliably get more modules that can be set into a 5GBASE-R mode: basically zero vendors advertise it.

[Some references](https://www.moduletek.com/en/application_notes/an_00085.html) indicate that modules based on the Broadcom BCM84891 will, by default, auto-detect 2500BASE-X or 5000BASE-X used on the host side interface. That may be easier to procure, since many module listings just plainly advertise the BMC84891.

storborg:
I've cleaned up and pushed my changes to get this working on the Acorn Mini and AC701:
https://github.com/enjoy-digital/liteeth/compare/feature/a7-5000baser...quadraturecat:liteeth:feature/a7-5000baser-working

This depends on the BIOS code I added in https://github.com/enjoy-digital/litex/pull/2563, but you could also just comment that out and use remote I2C commands.

The second commit is a 5/10GBASE-R PHY for Kintex-7 GTX transceivers. This is much simpler. The PMA is a little more straightforward because of the additional gearbox modes in the GTX, and obviously it avoids all of the SFP+ wrangling discussed in this thread when used at 10G.

storborg:
As an update in case anyone is following along, this transceiver:
https://www.amazon.com/dp/B07BHDJKRD

Looks like the most promising. It supports Rollball and can be configured to a 5GBASE-R host side interface.

As a bonus, it can also be configured to a 2500BASE-X host interface.

Unfortunately, it does *not* support 5000BASE-X.

enjoy-digital:
Thanks @storborg for the work/updates! Sorry I'm a bit busy to review/answer but will try soon.

storborg:
Hi @enjoy-digital, no problem, totally understand!

I have one more module on order left to test, a Broadcom BCM84891 based one.

PR #220, opened 2026-09-07
frontend/etherbone: send responses back to source port, fixes #214

### LINKED ISSUE #214: incorrect Etherbone UDP response port
**Describe the bug**

The Etherbone UDP response generated by the FPGA always uses the configured destination port instead of copying the UDP source port from the incoming request.

https://github.com/enjoy-digital/liteeth/blob/276c9e37fb4d92a5c0f30d39a51c20f42a59cf93/liteeth/frontend/etherbone.py#L64-L65

For example:

10.0.0.1 is the master
10.0.0.10 is the FPGA

request source port is random - 46045 in this case
request destination port is 1234

response source port is 1234 - correct, as configured
response destination port is 1234 - incorrect, should be 46045

```
No.	Time		Delta		Source		Destination		Protocol	Length	Info
12	6.720043526				10.0.0.1	10.0.0.10		eb			62		46045 → 1234 Len=20
15	6.720169459	0.000125933	10.0.0.10	10.0.0.1		eb			62		1234 → 1234 Len=20
16	6.720184024	0.000014565	10.0.0.1	10.0.0.10		ICMP		90		Destination unreachable (Port unreachable)
```

**Impact**

This breaks interoperability with standard Etherbone clients such as:
`eb-read`, `eb-write` tools from https://gitlab.com/ohwr/project/etherbone-core

**To reproduce**

Try reading 4 bytes from base address 0x00000008 of the FPGA using 
```
eb-read udp/10.0.0.10/1234 0x00000008/1
```

**Expected behavior**
The Etherbone UDP response generated by the FPGA should copy the UDP source port from the incoming request.

**Fix**
I will try to fix this and create a pull request.

storborg:
As described in #214, the Etherbone frontend currently sends response packets back to the fixed build-time port (1234) instead of using the original request's source port. This breaks compat with non-LiteX etherbone tools.

The fix here mirrors the handling of the source IP address, adding one 16-bit register. Test included.


enjoy-digital:
Thanks @storborg! This is merged.

PR #221, opened 2026-09-09
phy/pcs_1000basex: allow autonegotiation to be disabled

storborg:
This is needed for compatibility with some PHYs (e.g. Broadcom BCM84891) speaking SGMII or OCSGMII.

There's a lot of whitespace change here, but the substantive change is only to reduce the autonegotiation FSM to this when you pass `with_autoneg=False` to the PCS:

```python
self.comb += [
    self.align.eq(1),
    self.link_up.eq(~checker_error),
]
```


enjoy-digital:
Thanks @storborg! This looks fine. Can you just resolve current conflicts? and I'll integrate it.

storborg:
Actually, I think this may need some updates to be conceptually consistent with dc4a1c682418142889eca4ca53a94c7e87525fe1 from 2 days ago. My thinking is that passing `with_autoneg=False` should force `sgmii=True`, and maybe throw an exception if you pass a different value for `sgmii`.

@shenki any chance you could weigh in here? What SGMII PHYs led you to make this change?

storborg:
Related to, and maybe required for, #216

